PR #9, opened 2018-05-10
FCP HIPE: SSI Notation

dhh1128:


text/ssi-notation/README.md
```diff
@@ -0,0 +1,630 @@
...
+prominently in discussions about security and privacy, and our notation needs
+to be able to represent it carefully.
+
+The following crypto operations are recognized by the notation, without
+making a strong claim about how the operations are implemented. (For
+example, inline Diffie Helman and an ephemeral symmetric key might be
+used for the asymmetric algorithms. What is interesting to the notation
+isn't the low-level details, but the general semantics achieved.)
+
+* `anon_crypt(msg, recipient_pubkey)` -- Asymmetrically encrypt
```
vimmerru: For me "asymmetric" and "symmetric" words look like incorrect terms and can just confuse people. We don't use "asymmetric" encryption at all. "asymmetric" crypto is used for signing and DH. All encryption is always symmetric with MAC additions. 

anon_crypt means the following:

- get sender ed25519 key pair
- get the ed25519 public key of receiver
- call DH(sender_priv_key, receiver_pub_key, random_nonce) to produce SYMMETRIC key
- encrypt the message with SYMMETRIC key
- add MAC
- Return tuple (encrypted message, MAC, random_pub_key, random_nonce)

auth_crypt means the following:

- generate random ed25519 keys pair
- get the ed25519 public key of receiver
- call DH(random_priv_key, receiver_pub_key, random_nonce) to produce SYMMETRIC key
- encrypt the message with SYMMETRIC key
- add MAC
- Get tuple (encrypted message, MAC, random_nonce)
- Add sender public key and anon_crypt the whole message

swcurran: This is really useful to see.  I assume MAC is this - https://en.wikipedia.org/wiki/Message_authentication_code. 

Question - What key in the anon_crypt process is replaced with the "sender public key" in the last step of the auth_crypt process ("Add sender public key and anon_crypt the whole message")? The receiver knows the sender public key, so why is sent with the message?

swcurran:
@dhh1128 - I like this and definitely think it should be promoted.  I feel like the DIDDoc concept is missing and might simplify things. However, not quite sure where it should go - how much does it replace?  When I thought that a DID in Indy could only have one key and one endpoint, I thought it would be really useful (a DID implied that).

The other potential missing piece is where a DID is a microledger vs. a public ledger DID.  I'm not sure that's other than an implementation detail. A DID - micro- or public-ledger - can be resolved by those that need to resolve it, so where it is doesn't really matter. Either way, that resolution can fail - either because the Agent doesn't have access to the microledger or the DID doesn't exist - and that has to be handled.

text/ssi-notation/README.md
swcurran: Ah - I see.  Just add the sender public key to the message and then anonencrypt the whole thing with the same algorithm.

TelegramSam: Is there a reason the ciphertext (in auth_crypt) needs to be encrypted again? could you just anon_crypt the (MAC, randon_nonce, sender public key) portion, and send both?

kdenhartog: @vimmerru are authcrypt and anoncrypt reversed in step 3? I was thinking AnonCrypt uses random_priv_key and AuthCrypt uses sender_priv_key.

nage:
Without any further comments, I propose we merge this proposal, and will do so after the next maintainer call if no objections are raised.

swcurran:
I find the discussion that occurred in the comments about Anon/Auth is very useful and think it should be somewhere - add it to this or elsewhere?  Seems too specific to be a Notation spec. Or do I just save the URL to this PR :-) ?

As noted earlier, I think we should be including the concept of and using DIDDocs as a short form of encapsulating a lot of data about a relationship. For example, it encapsulates what is in a microledger.

dhh1128:
@swcurran I added a section about DID Docs. I feel this is ready to be merged now, though I'm sure we'll continue to find minor tweaks that we like, once we begin using the notation more broadly.

text/ssi-notation/README.md
```diff
@@ -277,6 +285,39 @@ dot P K at A to B” and “2 dot S K at A to B”. Here, `2` is known to
 belong to `A` because it takes `A`’s perspective on `A:B`--it would
 be equivalent but unnecessary to write `A.2.pk@A:B`.
 
+### DID Docs and DID References
+
+The mention of keys belonging to agents naturally raises the question of DID
+Docs and the things they contain. How do they relate to our notaiton?
```
swcurran: Typo - "notaiton"

swcurran:
I've reviewed the recent changes and am good with this being merged.

PR #16, opened 2018-06-01
 propose HIPE: Concurrency Improvement

vimmerru:
Libindy show provide better requests concurrency and performance in parallel requests
processing environment.

Note that this proposal can look a bit technical for HIPE, but some possible
options can significantly affect community and libindy developer experience.
So i suggest to discuss it this way.


dbluhm:
[Rendered text](https://github.com/vimmerru/indy-rfc/tree/feature/concurrency-improvment/text/concurrency-improvement)

dhh1128:
I am uneasy about this proposal. We are assuming a lot of responsibility for concurrency inside libindy, and making libindy fairly complex as a result. Some of the core things that libindy does are inherently asynchronous, such as interacting with a remote ledger. But I feel like crypto calls are inherently synchronous and should not be using a command object and a dispatch thread at all. By modeling such things as asynchronous, we are requiring consumers of crypto to model these synchronous operations as asynchronous with callbacks, and we are taking away from application code the ability to parallelize or serialize per its preference. In addition, I fear that the way we provide the concurrency inside libindy may limit its usefulness in certain embedded environments or on certain mobile platforms where concurrency assumptions may not fit the ones we've picked as our least common denominator.

There are two counter arguments to my worry, that I think are important:

1. If we don't do keep interfaces the same and build all the concurrency support into libindy in the way Slava proposes, we will have a lot of rewrite work. Slava makes this point in the HIPE, and I agree with it--but I don't think it ends all discussion. If we really are doing the wrong thing with the current interface, then we'd be far better off to rewrite now, before massive codebases get built on a flawed foundation, than to perpetuate a mistake. I think that the rewrite costs, while painful, would be bearable today. In a year, not so much.

2. Far more compelling in my mind is this counter-argument: Although crypto might be an example of inherently synchronous stuff in libindy, and although we might come up with a few other synchronous things as well, we might expect such functionality to rarely be consumed by itself. That is, most of the time we expect to be combining crypto and async comm. If this is true, that combining such calls is nearly always how the functionality is consumed, then keeping the crypto purely synchronous doesn't buy anybody a lot.

vimmerru:
In general idea of synchronous crypto looks ok for me too, but there is significant factor that makes our crypto endpoints very similar to pool requests endpoints. We don’t have just crypto. All crypto calls involves wallet and causes long file reading/writing or waiting on network sockets. So they are blocking io/cpu intensive operations. Modelling calls like this in a blocking API way will significantly complicate the most of applications. 

I am a bit sceptic about the point that it will simplify libindy. Usually providing blocking thread-safe API wihtout dispatcher thread is more complex task than providing non-blocking API.

dhh1128:
That's a good response, Slava. Thank you for thinking about my concerns. I would love hear what other people think about this topic. Are there members of the indy-sdk team, or coders out there in the community, that feel able to chime in?

nage:
As discussed in the Maintainers call and the WG call we had a call on this topic today and thw consensus was we are ready to move forward on this HIPE after a week comment period which starts today.  Please comment on this PR if you have blocking issues or feedback you'd like to see incorporated.

dhh1128:
Some of the Indy maintainers got together to discuss this interactively. We acknowledged that there's a tension here. If we take the approach advocated in this HIPE, almost nothing has to change in the APIs or wrappers, and client code only has to decide how big of a thread pool it wants. Libindy will become more capable of taking advantage of whatever concurrency is allowed by its host platform. And application developers will get certain concurrency benefits without thinking about them very much. Those are all good things. It would mean, for example, that TOB's use case where they're issuing 1.5M credentials will run drastically faster and will not block other indy operations.

On the other hand, this approach introduces some internal complexity into libindy that will have a carrying cost, and it requires that libindy run only on platforms that are capable of supporting sophisticated threading concepts. This may eliminate IoT things with embedded OSes, for example. Also, this approach is not very friendly to applications that want to be written for, and run as, synchronous processes. 

After a lot of discussion, we concluded that the benefits outweigh the drawbacks, so we will vote in favor of the HIPE. However, we want to do some research about what a synchronous API might look like because it may be desirable to wrap async libindy with a thin veneer that makes it friendly to synchronous use. That research is not a dependency of merging the HIPE, but should still be scheduled. Also, we concluded that for the benefit of the embedded/IoT use case (which relates to so-called "static agents"), we are probably going to have to commit to an ultra-simple library sharing some DNA with libindy, but lacking support for the ledger or wallets (the main things that must be asynchronous). This library would be synchronous and would probably have just the functions needed to package and encrypt messages, and to unpackage and decrypt them. Perhaps it would be indy-crypto, or perhaps something different...

We'll now move this HIPE to a final comment period.

nage:
Waiting period announced in Maintainers meeting has elapsed, please handle further issues as new PRs against this HIPE.

PR #17, opened 2018-06-26
approved HIPE: Agent Message Structure

dbluhm:
This HIPE is pretty basic but important for establishing some type of standard for agent developers.

It's my hope that more of the recent discussion around agents be formalized as HIPEs in the near future.


dbluhm:
[Rendered text](https://github.com/dbluhm/indy-hipe/tree/core-message-structure/text/core-message-structure)

text/core-message-structure/README.md
```diff
@@ -0,0 +1,63 @@
...
+  "id": "identifier/DID/nonce",
+  "type": "URN message type",
+  "message": {}
+}
+```
+
+- The id attribute is required and needs to be either the DID of the sender that is a pairwise identifier in an
+  established connection, a nonce used in establishing a connection, or another similar identifier used in a different
+  custom message exchange.
+- The type attribute is required and is a type string as outlined by a planned HIPE for message types.
```
pimotte: Why is the type required? 

There is something to say for having the message type inside the encrypted part of the message. This might something that's better addressed by the message type proposal, as it could have an "authcrypted message" type, where the specific type is then inside the encryption.

tmarkovski: Mostly because different message types have different encoding/encryption rules. Ex: connection request payloads aren't encrypted, response is anon-crypted and acknowledgement is auth-crypted. Entire base message is always anon-crypted during transport, so the data is only exposed at the agent service point.
You're right, moving the type inside the payload where it could be encrypted, means the base message will need to indicate payload encryption used.

text/core-message-structure/README.md
```diff
@@ -0,0 +1,63 @@
...
+The following `json`-like object is a representation of the proposed core message structure before being packaged and
+sent over the transport layer or after it is received through the transport layer and unpackaged. However, using `json`
+for messages is not necessarily part of this proposal.
+
+```json
+{
+  "id": "identifier/DID/nonce",
+  "type": "URN message type",
+  "message": {}
+}
```
tmarkovski: I would love to see destination/origination details added here. Using the current message description in an agent-to-agent scenario, I've always found the need to indicate either the destination DID for the message, or the origin DID or both. indy-agent repo added fields for this "aud" and "origin", but they're not documented here. I feel that need a discussion around this. Currently, I'm using similar format for the base message (proto syntax).
```proto
message Msg {
    string id = 1;
    string type = 2;
    string aud = 3;
    string origin = 4;
    bytes content = 5;
}
```

text/core-message-structure/README.md
dbluhm: @tmarkovski basically already said it but the type is used to know how to interpret the potentially encrypted message attribute. Additionally, as was already mentioned, the whole message is encrypted in transport so only the receiving agent is able to view the type.

text/core-message-structure/README.md
dbluhm: My experience in this is limited so I may be off but my understanding was that usually the "id" attribute contains the sender's DID and then the receiver looks up their DID and verkey for the relationship using the sender's DID (`pairwise.get_pairwise`).

Can you help me understand a case where sending the "aud" and "origin" attributes is required?

tmarkovski: Sure. A cloud agent acting as public endpoint on behalf on multiple edge (mobile) agents will need to know the message destination (aud) in order to route that message correctly. Edge agents in turn, will need to know the origin in order to find the pairwise connection to decrypt the message. This (origin) can be coded in the ID as you said (upon agreeing which message types contain ID as DID info, nonce, etc), though one challenge here is losing the message correlation identifier. Addressing information is mostly an issue when cloud agents are involved, which seems to be very common scenario that most mobile agents will rely on.

dbluhm: One solution that has been discussed for cloud agent/agency routing to the correct edge agents is having the edge agent essentially register DIDs/Identifiers that they want routed to them with the agent/agency that is acting as its endpoint. Here, then, the ID would remain as the sender's DID or a Nonce (depending on the message type). This idea is kind of discussed in the video by @dhh1128 on [agent-to-agent communication](https://drive.google.com/file/d/1PHAy8dMefZG9JNg87Zi33SfKkZvUvXvx/view).

I agree that this will be a very common scenario so if a change in this message structure can better accommodate it, let's do it. However, I think getting some more thoughts on this scenario from others would be great.

dbluhm: > though one challenge here is losing the message correlation identifier
I'm not certain what you mean here. If another identifier for the message is needed, it seems like it would belong in the "message" attribute with other content?

tmarkovski: Thanks @dbluhm - this makes sense. I watched the video as well. So the cloud agent will be able to figure out the destination of the edge agent it services, because the edge agent has associated certain senders to itself upfront. In the use case I was describing, however, the sender would instruct the cloud agent as to the destination of the message. I'd like to expand on this discussion on the trust assumptions on cloud agents. How does Alice prove that certain key/did pairs A:AB and B:BA are associated with each other?
If Bob would designate the destination audience (A:AB) himself with origin B:BA and Alice can prove she owns A:AB (Bob, too, can prove he owns B:BA), then the cloud agent at least has both source confirming they own the keys in question.
But if only Alice is to be trusted that A:AB and B:BA are associated (in scenario when Alice registers the association upfront), then the cloud agent can't be sure the association in Bob's domain is true.
I hope this makes sense.

tmarkovski: More thoughts on this: suppose a scenario where Alice send a connection request to Faber, which contains her DID. Faber, generates new DID for connection with Alice and sends the connection response to Alice's cloud agent. At this point, Alice doesn't know Faber's new DID, so she can't register this connection with her cloud agent. The connection response messages seems undeliverable. Does Alice register the request nonce instead with her cloud agent? If yes, should the cloud agent be aware of nonce vs did association? How complex should the nonce be to guarantee uniqueness?

dbluhm: > In the use case I was describing, however, the sender would instruct the cloud agent as to the destination of the message.

I think this is giving away too much information to your connections. Additionally, this requires informing your connections of the internal structure of your domain, giving away correlatable pieces of information.

> Alice send a connection request to Faber, which contains her DID. Faber, generates new DID for connection with Alice and sends the connection response to Alice's cloud agent. At this point, Alice doesn't know Faber's new DID, so she can't register this connection with her cloud agent.

I think in this case, Alice would generate a nonce that Faber would then use as the identifier in the response. Alice does register the request nonce with the cloud agent. And yes, I think it would be valuable for Agencies/Cloud Agents to know the difference between a nonce and DID when they are registered for routing.

I think this conversation is descending into more of a discussion of Agent-to-agent communication which probably belongs in a different HIPE. However, again, if we can make a core message structure change that will better facilitate communication, then let's do it, but I think it would be good to get more comments on this.

tmarkovski: I think this structure covers all scenarios we're currently working with. Thank you for the clarification on did/nonce registration with the cloud agent, this takes away the need to address each message by the sender.

text/core-message-structure/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+The following `json`-like object is a representation of the proposed core message structure before being packaged and
+sent over the transport layer or after it is received through the transport layer and unpackaged. However, using `json`
+for messages is not necessarily part of this proposal.
+
+```json
+{
+  "id": "identifier/DID/nonce",
+  "type": "URN message type",
+  "message": {}
```
ryanwest6: This looks like a clear and concise way of sending messages. My only comment would be to perhaps rename the inner "message" field to be "content". This would resolve possible ambiguity between the overall message json, and the currently named "message" field which it contains.

dbluhm: I agree that having the inner content field named "message" can be confusing or redundant. In code for instance, you might commonly name the message received "message" then accessing the message attribute would be `message.message`. Something like `message.content` is feels nicer.

text/core-message-structure/README.md
kdenhartog: I agree with this message structure for now, and that in the future we should be cognizant of the idea that the frequency of messages combined with the message type can be used to fingerprint who the identity is that is being communicated with. For example, a person who is participating in specific activities through their personal communications will be easier to identify because they are sending more of a specific message type to many different people and they are therefore, highly active in a particular activity. However, I don't think finding the perfect solution should prevent us from moving forward with a good enough solution, so I'm willing to move forward with this proposal until a realistic problem is presented and we can address it then.

swcurran: This looks good. I think that we need a HIPE on what "transport layer messaging" means as that is a really important part of this HIPE.  It could be in this HIPE, since this is the definition of core, or as a separate one, but either way, we need to make it clear what that means.

blhagadorn: One of the unfortunate parts about it is that the high activity and lower user count to begin with, at least in our case, won't be helped until we get more of a network to create noise.  However, this is a very great solution to move forward, excellent work!  
One option, until a realistic problem arises, is for those of us who don't want our data extremely corollary to start with (high spikes of messaging with few users), is to potentially (internally) set some helpful implementation suggestions that might mitigate that problem in the beginning or until a better solution comes out.  A few ideas that come to mind would be smoke messages, timing queues (only release from the application layer every 15 seconds or at a random time intervals), and general random and unpredictable activities at the expense of a little bit of a rough user experience to create noise.  Curious if anyone can think of how that actually might be detrimental, however.
Sorry and please correct if this is off topic or not a part of the HIPE guidelines.

saholman:
Sorry to come late to the game, but I don't think we can get rid of the `id` field.  The agency needs a single field to look at in order to route the message, because otherwise the agency has to know about every message type and how to route messages of that type.  I think that we keep the previous message structure of ```

saholman:
We can call the `message` field `content` or `payload` or whatever, but its important to keep the payload of the message separate from the id used for routing and the type used so the agent knows where to send the message.  The data contained in the `message` field is what is sent to the handler of the message for a specific type.

dbluhm:
@saholman I agree that some type of identifier will absolutely necessary for nearly all communications. If we let the type dictate all other attributes in a message, though, it allows attribute names that may be more intuitive for a given type. For instance, using "to" or "aud" or whatever in the place of "id" when the identifier is used to route to another agent.

The necessity of these identifiers and their names will be defined as message types are defined.

saholman:
@dbluhm If the type indicates which field contains the routing information, then the agency would have to know about every possible type of message in order to route them.  Having a standard `id` field in the outer message allows the agency to look in a single place for the routing identifier, and the `to` or `aud` attributes can be inside the encrypted `content` section if they are still needed by the message handler.

saholman:
After the connection, it seems like every other message could use the `their_did` for routing messages.  This is convenient because the `their_did` is the same DID used to look up a pairwise in indy-sdk.  Thoughts?  If that were the case, then the standard could say that the only thing that would be in the `id` field (or whatever we want to call it) after the connection would be the `their_did` regardless of the message type.

text/core-message-structure/README.md
mhailstone: @kdenhartog I just want to emphasize here that the only entities who would be doing the data gathering would be the endpoint service or agencies. All the data over the wire would be encrypted and not readable so the message type would not be collected for process outside of the endpoint/agency service. I think we need to define guidelines on what is trustworthy for an agent/agency to know and what information they should not be allowed to know and protected from knowing through the protocol we define.

dhh1128:
I think we need to distinguish between the need to describe the type of a message in two different contexts:

- describe the type from *within* the message
- describe the type from *outside* the message

The type that's visible inside a message is guaranteed to be there even if we divorce the message from its transport or context and save just the raw bytes of the message to disk. It is important to have this, so that messages can be self-contained. This is why JPEG files have an internal header that identifies them as JPEGs, PDF files have an internal header that identifies them as PDFs, and so forth.

However, a type declared internally is of no use to something that sees a message payload only as an encrypted blob of bytes. So we also need to declare type in a way that's external to the message itself, such that things that are processing the message (or routing it to other things to process) have the option of learning something about what they are handling.

The current proposal feels to me like it conflates these two concerns, placing a type declaration outside the message content but inside some other container that must unify the two.

The original A2A design called for a type declaration to be a message type of its own. You could take any other message and make it a payload of a type message. This is very close to what this proposal recommends, except that in the original view, the thing you were looking about when you were done wasn't a message of arbitrary type but guaranteed to have type as a field--but rather a "type" message that had another message as its payload. If we could think about it that way, I think we'd be better off.

TelegramSam:
This could use two minor changes:
1. Specify that attributes with an @ prefix are reserved.
2. Use `@type` instead of `type`

PR #18, opened 2018-07-06
Propose HIPE: Connection Protocol

ryanwest6:
https://github.com/ryanwest6/indy-hipe/blob/master/text/connection-protocol/README.md


text/connection-protocol/README.md
```diff
@@ -0,0 +1,272 @@
...
+    "type": connection_offer
+    "content": {
+        "endpoint": {
+            "did": A.endpoint.did
+            --- or ---
+            "uri": A.endpoint.uri
+            "verkey": A.endpoint.vk
+        }
+    }
+}
```
mhailstone: We've learned a lot since our original proposal in the Google document. Here is a new proposal:
```
{
  "type": <connection_offer_urn>
  "message": {
    "offer_nonce": offer_nonce,
    "endpoint": {
      "did": A.endpoint.did,
      --- or ---
      "location": A.endpoint.location
      "verkey": A.endpoint.vk
    }
  }
}
```
We don't need the `id` attribute any more as a required attribute at the core message structure. The `type` attribute (which contains the message type) will dictate what attributes (other than `type`) that need to exist. I like the convention of `message` instead of `content`, but that really doesn't matter. We just need to decide. I'm moving the `offer_nonce` to be inside the inner message because the inner message is not encrypted at all anyway, and we don't have to repurpose the `id` attribute to contain the nonce. I intentionally avoided using `uri` as the value of the endpoint because I wasn't certain what the structure of the endpoint should be based on the variation of protocols that may need to be supported in the future. I left it as a generic `string` value. If we decide to restrict it as a `uri` I think we should propose a structure for the `uri` pattern to follow as the value (similar to the value of the message type).

text/connection-protocol/README.md
```diff
@@ -0,0 +1,272 @@
...
+```
+
+#### Attributes
+
+* The `id` attribute of the base message is required and is a nonce used to correlate the connection request.
```
mhailstone: This has changed. The `id` attribute is now not required, but other attributes may be introduced based on the definition of the specific message type structure.

text/connection-protocol/README.md
```diff
@@ -0,0 +1,272 @@
...
+```
+{
+    "id": offer_nonce (optional; dependent on scenario)
+    "type": connection_request
+    "content": {
+        "did": B.did@B:A
+        "verkey": B.vk@B:A
+        "endpoint": B.endpoint
+    }
+}
```
mhailstone:
```
{
  "type": <connection_request_urn>,
  "message": {
    "offer_nonce": <offer_nonce>,
    "request_nonce": <request_nonce>,
    "did": B.did@B:A,
    "verkey": "B.vk@B:A,
    "endpoint": B.endpoint
  }
}
```
The `offer_nonce` attribute needs to exist as an optional attribute in case the connection request is in response to a connection offer. The `request_nonce` attribute is required so that the connection response message can similarly correlate with a previously sent connection request. The connection request is again not encrypted (other than the transport, but that decryption has already happened by this point in the process), so the `id` attribute is not required and putting all the data inside the `message` object encapsulates the message data nicely.

text/connection-protocol/README.md
```diff
@@ -0,0 +1,272 @@
...
+
+[The entirety of this message (and all other messages in the connection process) is anon-encrypted using the endpoint
+information that Alice sent in the connection offer](#unresolved-questions), ensuring that all information is encrypted
+in transport and no correlateble data can be leaked.  At this point, not enough information has been exchanged for Bob
+to be able to encrypt the content for Alice only in the context of their relationship (Bob does not yet have a
+verification key for Alice in the Alice to Bob relationship).
+
+#### Attributes
+
+* The `id` is required when a connection offer was used to initiate the connection establishment process and is the
```
mhailstone: This has changed. The `id` attribute is now not required, but other attributes may be introduced based on the definition of the specific message type structure.

text/connection-protocol/README.md
```diff
@@ -0,0 +1,272 @@
...
+
+```
+{
+    "to": A.did@B:A,
+    "type": connection_response,
+    "content": {
+            "did": A.did@A:B,
+            "verkey": A.vk@A:B
+    }
+}
```
mhailstone:
```
{
  "type": <connection_response_urn>,
  "to": <request_nonce>,
  "message": {
    "did": A.did@A:B,
    "verkey": A.vk@A:B
  }
}
```
I'm putting the `request_nonce` in the `to` attribute because the receiver of the connection response message uses the nonce to route who it's for as well as for the reason that the `message` object is (currently) anon-encrypted and Base64 encoded, so the data in the `message` object is not able to be read by the endpoint process that is reading the message.

text/connection-protocol/README.md
saholman: We can't get rid of the id field!  The id field is necessary for routing within the agency.  The agency needs a single attribute it can look for to route the message to the agent.

dbluhm: The ID is certainly necessary for routing but I do agree that where it is for the connection offer specifically might not make the most sense. Especially since the connection offer is out of band communication. I'd like to hear other thoughts on this.

> The type attribute (which contains the message type) will dictate what attributes (other than type) that need to exist.

:heavy_check_mark: 

>  I like the convention of message instead of content, but that really doesn't matter.

Consider the case where in code you name the message received `message`. If we used message to name the inner payload, we would then get `message.message` in our code. This just felt kind of funny to us. `message.content` feels a little more natural. But again, not incredibly important by any means.

>  I intentionally avoided using uri as the value of the endpoint because I wasn't certain what the structure of the endpoint should be based on the variation of protocols that may need to be supported in the future. I left it as a generic string value. If we decide to restrict it as a uri I think we should propose a structure for the uri pattern to follow as the value (similar to the value of the message type).

We had a precedent in the SSI Notation HIPE by @dhh1128 for calling the actual string pointing to an endpoint the `uri`. I went along with that because `uri` is probably the most general term we could use to describe some sort of remote location. It might be something like `http://example.com` or it could be something that isn't even defined yet.

text/connection-protocol/README.md
dbluhm: Please see my comment above about the id field.

text/connection-protocol/README.md
dbluhm: When the offer_nonce is needed, I think it ought to be in the outer-message just like other messages where the identifier used to route must be in the outer-message. My reasoning being partially just for consistency and the other part being that I think finding some way to encrypt the inner content of a connection request is valuable (so we don't have to trust the agency). At that point, the nonce would obviously have to be in the outer-message.

text/connection-protocol/README.md
dbluhm: See my comment above.

text/connection-protocol/README.md
dbluhm: Why use a request nonce as the routing identifier if the party that sent the request already knows did@B:A and the party sending the response also knows the did@B:A?

text/connection-protocol/README.md
ryanwest6: This discussion actually relates somewhat to one of the unanswered questions we posed at the bottom. Because the message type dictates the other attributes present in the message structure, we may not necessarily even need to have an inner `content` structure at all in this one case.

text/connection-protocol/README.md
ryanwest6: I'm trying to see the benefit of a `request_nonce` to identify the request and the response. While I understand its proposed purpose, it seems like it would be just as possible to correlate the request with the response based on the pairwise DID and verkey specifically generated for this relationship. Nobody could try to forge a false response unless they could obtain that DID and verkey, in which case they could also obtain the nonce. Am I missing another reason?

text/connection-protocol/README.md
saholman: @dbluhm I agree that the id doesn't add any value, and doesn't really make a lot of sense in the connection offer.  I agree that `message.message` is a little awkward, and I don't think it really matters what we call it.  @ryanwest6  I think the inner `content` structure makes it cleaner because the type and id are no longer necessary once the message gets to the specific handler and the right agent, so the agent can simply pass that inner `content` structure to the message handler.

text/connection-protocol/README.md
saholman: @ryanwest6 I believe the request nonce's purpose is not to correlate the request and response, but rather to validate that it's still talking to the same person.  Agent A generates a nonce, and it makes sure that same nonce comes back in the connection response.

text/connection-protocol/README.md
saholman: @dbluhm See my comment above about the request_nonce.  This comes from the original python getting started guide.  @mhailstone I don't think the request_nonce should be used as the routing identifier either.  The request_nonce has a different purpose, and making it "multipurpose" might be confusing.

text/connection-protocol/README.md
ryanwest6: Thanks for the response. I understand that a nonce could be used for that purpose, but what I was trying to say at the end of my last sentence was this: If a request is sent from Bob to Alice, then only Alice has the pairwise DID (and its corresponding verkey) that Bob just generated for the Bob-to-Alice relationship. So nobody could impersonate Alice sending a response unless they gained access to that connection request message and learned both the unique DID and the verkey. So there is no need to worry that the person who responds is not Alice. (If they did gain access to that message, then they would also gain access to `request_nonce` and could impersonate her anyway.)

text/connection-protocol/README.md
saholman: @mhailstone I don't think the request_nonce should be used as the routing identifier.  The request_nonce has a different purpose, and if it is in the clear (unencrypted) then the validation of the request_nonce could be spoofed.  At the same time, we have to figure out how the connection response will be routed.  In all other message types, messages should be routed using the "their_did" as the id because a pairwise is looked up by the "their_did" of the relationship in the indy-sdk.  Using our own did here only might also be confusing, but its probably our best option.  That is what we are currently doing in the node.js indy-agent implementation.

text/connection-protocol/README.md
saholman: @ryanwest6 I think your right.  This idea comes from the original python getting_started guide and we've carried it through since then.  Unless someone else has a really good reason for keeping it, I have no problem getting rid of it.  Take a look at my comment below about routing the connection response too.

saholman:
So the question we need to answer is whether or not we need a `request_nonce`.  If we get rid of the `request_nonce`, then we will use the `my_did` for routing the `connection_response` where as most all other messages will use the `their_did`.  As @ryanwest6 pointed out, I don't know that the `request_nonce` is adding any real security.

dbluhm:
As per discussion in the Indy Maintainers call this morning, we're going to rework this HIPE to exclude any details on transport. These details will instead be included in a separate HIPE on transport messages. @swcurran has volunteered to write up that HIPE.

text/connection-protocol/README.md
mhailstone: But, we are not sending the requestor's (B) DID and verkey in the "connection response" message. B does not know which of their DID/verkey combination to associate (pairwise) this new A's DID/verkey it is receiving.

ryanwest6:
Updated this HIPE by removing transport layer details (such as `id`/`to`). They will be covered in the other HIPE instead. Also removed references to anoncrypt and authcrypt

dbluhm:
As suggested by @dhh1128, I think using terminology that can be easily generalized such as the "negotiate message flow" would be a valid and valuable change here -- especially if its usage becomes more widespread across other SSI platforms.

text/connection-protocol/README.md
```diff
@@ -0,0 +1,244 @@
...
+    {
+        "type": connection_offer
+        "endpoint": {
+            "did": A.endpoint.did
+            --- or ---
+            "uri": A.endpoint.url
+            "verkey": A.endpoint.vk
+        }
+    }
+    ```
```
tmarkovski: I wanna say the messaging format should conform to some schema/protocol, instead having content driven format. This would make it easier to maintain, implement and transform the messages. Furthermore, in strongly typed environments, having loose structure adds a lot of complexity. This may work well in javascript runtimes, but it's not as convenient in java, dotnet, swift, etc. As such, I think keeping the `message` or `content` field fixed is the better way to go.

text/connection-protocol/README.md
```diff
@@ -0,0 +1,244 @@
...
+            "verkey": A.vk@A:B
+    }
+}
+```
+
+#### Attributes
+
+* The `type` attribute is a required string value (following the structure outlined by a future HIPE on message
+  types) and denotes that the received message is a connection response.
+* `content` is anon-encrypted using the receiver's verification key sent in the connection request message. It includes:
```
tmarkovski: There's no reason why the `content` can't be authcrypted at this point. The creator of the Response has both keys to auth-crypt, and the receiver of the Response only needs their key to auth-decrypt.

dhh1128:
I wanted to share a spreadsheet that I worked up, that analyzes connections as a state machine. I think it is largely compatible with Ryan's vision, but there are some subtleties worth discussing. Specifically: I have eliminated the ACK and ACK ACK. This is because I think any messages that flow after the connection is made accomplish the same purpose, without us allocating a special message type and special message send for them (e.g., if I immediately sent a proof request after establishing a connection, the same knowledge would be gained). Also, I think the ACKs tell us less than we'd suppose, since any connection can decay as soon as a message is sent. Also, I'm assuming the notion of negotiation back-and-forth (request and offer and request and offer...) until both parties are satisfied, even though I don't imagine most connections will negotiate much. And I'm accepting the reality that each party may have a different view of the connection's state, so I'm only trying to model the view on one side.

Maybe we can discuss in next week's agent meeting, or I can record a 10-min video where I walk through my thinking...

https://docs.google.com/spreadsheets/d/1RLJhhlWCUBYpKn18S5BGi7HH9MfGiZdHBxKjsJavoMs/edit?usp=sharing

dhh1128:
I recorded an explanation of what I intended with the spreadsheet:  http://bit.ly/2nhQMRC. When can we discuss interactively?

dbluhm:
I posted this on Rocket.Chat as well:

@dhh1128  I'm looking through your spreadsheet about connections and watching your accompanying video as well. Thanks for putting those together. You focused more on the transitions from one state to the next rather than what is transmitted with each step so I wanted to paraphrase what I interpreted and get your feedback to make sure I was on the same page:

#### Alice initiates a connection with Bob:
1. Alice sends an out of band invitation to Bob.
2. Bob accepts the invitation, proposing a DID and key to be used. (I'm not sure I understood this correctly but that is what I interpreted from your assumption: "The invitation phase ends when the initiating (inviting) party is contacted and is able to transmit a Connection Request that is A2A encrypted with the DID+key they propose to use.")
3. Alice sends a Connection Request to Bob using the information provided by Bob when he accepted the invitation.
4. Bob sends a "Connection Outcome" to Alice, notifying of acceptance or rejection. (Is anything else transmitted here?)

smithbk:
In thinking of the connection, issuance, and verification flows as state machines. I'd like to suggest the creation of generic CRUD type messages.  I don't think this is very different, but does make the message types more canonical and so reusable for other flows.  For example, here is the puml for what I was thinking with create and update messages. 
`@startuml

actor Alice as A #white
actor Bob as B #white


rnote over A #white
{
    "type": "create",
    "object": "connection",
    "id": "<connection-id/nonce>",
    "state": "offer",
    "offer": {
        "endpoint": {
            "did": A.endpoint.did
             — or —
            "uri": A.endpoint.uri
            "verkey": A.endpoint.vk
        }
    }
}
end note

A -> B: **1.** A sends **Connection Offer** in\nperson or securely over internet
||35||


rnote over B #white
{
    "type": "update",
    "object": "connection",
    "id": "<connection-id/nonce>",
    "state": "request",
    "request": {
        "did": B.did@B:A
        "verkey": B.vk@B:A
        "endpoint": B.endpoint
    }
}
end note

B -> A: **2.** B sends **Connection Request** message
||35||


rnote over A #white
{
    "type": "update",
    "object": "connection",
    "id": "<connection-id/nonce>",
    "state": "response",
    "response": {
            "did": A.did@A:B,
            "verkey": A.vk@A:B
    }
}
end note

A -> B: **3.** A sends **Connection Response** message
||35||


rnote over B #white
{
    "type": "update",
    "object": "connection",
    "id": "<connection-id/nonce>",
    "state": "connected"
}
end note

B -> A: **4.** B sends **Connection Acknowledgement** 
||35||


rnote over A #white
{
    "type": "update",
    "object": "connection",
    "id": "<connection-id/nonce>",
    "state": "connected"
}
end note

A -> B: **5.** A sends **Connection Acknowledgement** 
||35||


A <-> B: A & B can now send auth-encrypted messages

@enduml`

text/connection-protocol/establishing_connection.puml
```diff
@@ -0,0 +1,60 @@
...
+Bob --> Alice: request
+rnote right #white
+anon_crypted(endpoint_key, {
+    @type: "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/routing/1.0/forward_to_key",
+    key: connection key (one time use),
+    content: {
+        @type: "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/connections/1.0/request",
+        key: connection key (one time use),
+        content: auth_crypted(connection key, {
+            did: // did on a public ledger
```
trthhrtz: @kdenhartog documentation needs to be updated when only DID is necessary or full did doc is.

text/connection-protocol/establishing_connection.puml
```diff
@@ -0,0 +1,60 @@
...
+
+Bob --> Alice: request
+rnote right #white
+anon_crypted(endpoint_key, {
+    @type: "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/routing/1.0/forward_to_key",
+    key: connection key (one time use),
+    content: {
+        @type: "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/connections/1.0/request",
+        key: connection key (one time use),
+        content: auth_crypted(connection key, {
```
trthhrtz: @kdenhartog change this to anon_crypt

text/connection-protocol/establishing_connection.puml
```diff
@@ -0,0 +1,60 @@
...
+       endpoint uri
+       endpoint key
+    * connection key (one time use)
+end note
+
+Bob --> Alice: request
+rnote right #white
+anon_crypted(endpoint_key, {
+    @type: "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/routing/1.0/forward_to_key",
+    key: connection key (one time use),
```
trthhrtz: @kdenhartog ?

text/connection-protocol/establishing_connection.puml
```diff
@@ -0,0 +1,60 @@
...
+            did_doc:
+        })
+    }
+})
+end note
+Alice --> Bob: response
+rnote left #white
+anon_crypted(endpoint_key, {
+    @type: "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/routing/1.0/forward",
+    to: "ABC",
```
trthhrtz: did # keys or just verkeys, or just did? @kdenhartog

text/connection-protocol/establishing_connection.puml
```diff
@@ -0,0 +1,60 @@
...
+    content: {
+        @type: "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/connections/1.0/response",
+        to: "did:sov:ABC",
```
trthhrtz: did # keys or just verkeys, or just did? @kdenhartog

mikelodder7:
For an introduction message, forward secrecy could be achieved by just doing a Diffie Hellman exchange as part of the sequence. Why can't we simplify the connection protocol to just be:

A -> B: One-Time-Public Key (OTK-A), Reverse-URL
B -> A: One-Time-Public Key (OTK-B), Enc(K, DID, Identity Key)
A -> B: Enc(K, DID, Identity Key)

K is computed by B using OTK-A and his private material and
A using OTK-B and her private material.

This wouldn't provide mutual authentication and is subject to MITM but so is every introduction handshake. If this exchange happens offline (QR code scanning, code typing) then this is very unlikely to be the case. Microledgers could also be included as part of this exchange. Other data can be exchanged as well but the point is that we don't need authcrypt or anoncrypt for the intro handshake.

If party B has a well known public key say in their DID Doc on the ledger, then A can authenticate B.

The only information sent in the clear in the first message is the public key, the reverse url, and message type like offer. The second message encrypts everything but the public key.
The third message is all encrypted. You could just use regular vanilla encryption using any algorithm.

dbluhm:
I propose we close this HIPE in favor of the simplified and more secure protocol proposed by @mikelodder7 #46.

brentzundel:
Has this HIPE been superseded by #54 ?

dhh1128:
Closing this PR. It contains many good ideas, but it is being superseded by PR #54 .

PR #30, opened 2018-08-03
FCP HIPE: Message Threading

dbluhm:
Signed-off-by: Daniel Bluhm <daniel.bluhm@sovrin.org>


swcurran:
Looks good!

My first thought in reading this were that "mid" messages were correlated - request/response - but that is not right?  I could send multiple messages (mid 1, 2 and 3) without the other side responding, right?

Should there be any error handling?  For example, I get a message 3 from the other side without getting the message 2?  I'm guessing that should be at the higher level, but perhaps there should be an explicit thread-family message (vs. a decorator) that says "Hey, I didn't get message 2".

There should be some cleanup (I think) of the JSON - the @thread aligned with the @type and the fields indented and aligned with each other?

dbluhm:
@swcurran Thanks for bringing up the indentation issue.

Yes, I believe it is intended that the `mid`s of both sides are not dependent on each other. So you could have several messages sent from one side (1, 2, 3) without response.

As far as error handling goes, I think we would want to leave that to higher levels. For instance, there might be some contexts where a single dropped message isn't significant enough to warrant emitting an error or can be safely ignored. It seems like the message family will have to define whether it should emit an error on a dropped message. Doing it that way also allows the message family to dictate the type and attributes of that message which would probably be more helpful than a generic message thread error.

That being said, I'm glad to hear other peoples thoughts on error handling.

dhh1128:
This is related to @swcurran 's question about error handling. What happens if A expects B to say something at a certain point, and B is silent? A begins to wonder what B is up to, and if B needs a nudge. A wants to send a messaging saying (effectively), "Hey, here's how far we got in our interaction, as far as I can tell. Did you say something while you had your mute button on? If so, can you please resend beginning after your message id"? Now, this "yoohoo, I'm waiting; please (re)send" message is probably not something we need to document in this HIPE--but when B receives such a message, how are the resent messages numbered? If messages 4-6 had been sent before, but never received, and B resends, are they still numbered 4-6? Or are they numbered 7-9, but marked as replacements for messages 4-6? I think maybe the latter, but that means there's another possible field in @thread. Maybe "resend-of-mid"? Or maybe we have a field called "send-attempt" and reuse the message ids, but mark a resent as "send-attempt: 2" or similar?

This makes me wonder if we need the ability to prune/orphan part of a conversation (e.g., message 7-8 occur based on certain assumptions about the other party's state; then the other party's state changes, and we want to abandon what we said in messages 7-8. Imagine a negotiation where we want to abandon an earlier offer once we know more about what the other party values; we don't want them, late in the game, to go back to an earlier offer and suddenly accept it when we've evolved way past it). Not sure how this relates to threading, but I've got a vague worry that it creates message trees instead of threads...

dhh1128:
Regarding my last comment, this may be a place to do some research about message threading in chat protocols...

swcurran:
I like the idea of including the optional "send-attempt" (if not sent assume "1").  I think that could be used as well for the "are you on mute" message. A participant that has not heard from the other conversant in a while doesn't send an error, but rather resends their last message (same mid) with the "send-attempt" sent to one higher.

I'd say we leave the vague worry alone for now and not worry about it until we have some concrete examples of what we need.

text/message-threading/README.md
```diff
@@ -0,0 +1,210 @@
...
+{
+    "@type": "did:example:12345...;spec/example_family/1.0/example_type",
+    "@id": "98fd8d72-80f6-4419-abc2-c65ea39d0f38",
+    "@thread": {
+        "thid": "98fd8d72-80f6-4419-abc2-c65ea39d0f38",
+        "pthid": "1e513ad4-48c9-444e-9e7e-5b8b45c5e325",
+        "seqnum": 2,
+        "lrec": 1
+    },
+    "msg": "this is my message"
```
tplooker: Im guessing the message isn't nested inside here, instead the attributes defined by the @type will be listed in this outer message. Perhaps change this property from "msg", to "example_attribute" like the JSON above, if that is the case?

dhh1128: Exactly right. Good suggestion!

text/message-threading/README.md
```diff
@@ -0,0 +1,144 @@
...
+The thread object can be associated with a message in one of two way.
+
+#### Implicit Threads
+
+Threads reference a Message ID as the origin of the thread. This allows _any_ message to be the start of a thread, even if not originally intended. Any message without an explicit `@thread` attribute can be considered to have the following `@thread` attribute implicitly present.
+
+```
+"@thread": {
+    "tid": <same as @id of the outer message>,
+    "mid": 0
```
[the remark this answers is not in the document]
dbluhm: I believe @TelegramSam fixed this with his additions.

text/message-threading/README.md
```diff
@@ -0,0 +1,210 @@
...
+can include a Parent Thread ID (`pthid`). This signals to the other party that this
+is a thread that is branching off of an existing interaction.
+
+##### Nested Example
+As before, Alice is an issuer and she offers a credential to Bob. This time, she wants a bit more information before she is comfortable providing a credential.
+
+* Alice sends a CRED_OFFER as the start of a new thread, `@id`=98fd8d72-80f6-4419-abc2-c65ea39d0f38, `seqnum`=0. 
+* Bob responds with a CRED_REQUEST, `@id`=&lt;uuid2&gt;, `thid`=98fd8d72-80f6-4419-abc2-c65ea39d0f38, `seqnum`=0, `lrec`=0.
+* **Alice sends a PROOF_REQUEST, `@id`=&lt;uuid3&gt;, `pthid`=98fd8d72-80f6-4419-abc2-c65ea39d0f38, `seqnum`=0.**
+* **Bob sends a PROOF, `@id`=&lt;uuid4&gt;, `thid`=&lt;uuid3&gt;,`seqnum`=0, `lrec`=0.**
```
swcurran: Sorry to raise this again, but wouldn't the pthid from Alice's Proof Request be carried into Bob's Proof response?  This was addressed already from a previous comment, but I thought the the pthid would continue to be carried in the subsequent messages but it's not in this example.

dhh1128: I believe the thinking is that you only need the parent to be identified when you initially fork. The logic is that if you can resolve to the head of the current thread, then you can also find out that there's another message under which the current thread is a subthread. If, on the other hand, you can't get back to the message at the start of the current thread, then there's near zero chance that you could do anything useful with the knowledge of a parent thread. Also, most participants in the current thread probably don't care about the parent thread.

So that's a coherent viewpoint, but I don't think it's the only valid one. We could carry `pthid` around, and it would have the advantage of pointing everyone back to a higher context until the thread resolves. The only argument against that, as far as I can tell, is verboseness.

On balance, i kind of like the less verbose form. But I could be convinced either way.

dhh1128:
I just realized that, despite having worked on this HIPE myself and having thought very carefully about it, I was using the `seqnum` construct wrong. It is an easy mental error, because it's SO natural to assume that seqnum is not sender-specific but rather a sequence number with the thread as a whole. Because of this, I am feeling more opinionated about renaming `seqnum`. As I suggested in my previous comment, something with "idx" in it would fit with the zero-based semantics. But now I also want the name to clarify its sender-specific nature. So how about `myidx`?

TelegramSam:
I'm not sure `idx` is any less confusing (in different ways, possibly) than `seqnum`. Not opposed to a rename, but I don't think we have it yet.

TelegramSam:
@dhh1128 What about `senderseqnum` as a replacement for `seqnum`? Or perhaps `myseqnum`? Moving to `myidx` feels like we are trading one confusion for another.

swcurran:
> @dhh1128 What about `senderseqnum` as a replacement for `seqnum`? Or perhaps `myseqnum`? Moving to `myidx` feels like we are trading one confusion for another.

I like "myseqnum". Makes it pretty obvious - more so than the other suggestions.  "senderseqnum" would also be fine.  I don't get "idx" at all (index?  if so, use "index" :-) ), and even so, it would need the "my" or "sender" prefix to be useful.

dhh1128:
I like "myseqnum" moderately well. I like "myseqindex" or even just "myindex" better, because they *also* are better for my start-counting-at-zero concern. But at this point, any of the field names that have "my" would pass my part for merging the HIPE. So @dbluhm, I say pick one and let's get it merged.

dhh1128:
@dbluhm and @swcurran and @TelegramSam : I have raised a PR against Daniel's fork that, if accepted, would update this HIPE proposal as follows:

1. Change `seqnum` to `myindex` (this turned out to involve more lines of changes than I expected)
2. Remove the idea of `lrec` that can have 2 different forms (one singular, one plural). Instead, standardize on plural `lrecs`. I did this because I think it will be better for JSON libraries to convert JSON to native object form in a consistent way.

TelegramSam:
I love lrecs being standardized.

I'm not quite settled on myindex. I didn't have a problem with seqnum, but as long as we are looking to change it, I wouldn't mind there being better naming similarity between the seqnum/myindex and lrecs, the former being the message numbers, the latter being the last ones seen.

PR #31, opened 2018-08-06
propose HIPE: Relationship State

devin-fisher:
Currently a work-in-progress!

The architectural theory is mostly missing. Please see [this document by Jason Law](https://docs.google.com/document/d/1dFmQ_5NlAIrCYx7AYvvEeYlyvzv-2u3cpQqWTadJbPA/edit) that is the bases of what will go into this portion of the hipe. 

Technical details are rough but flushing out. They are coming from a POC that evernym is researching. 
([See POC - code only](https://github.com/lovesh/indy-sdk/tree/rsm))


swcurran:
I see the issues behind this HIPE as being a significant roadblock to making progress with A2A.  What do we need to do to break the blockage?  The issues I see:

1. Indy-SDK (at least, perhaps Indy-Node as well) does not support the elements of a DIDDoc - multiple keys, multiple endpoints, multiple authorizations, and endpoints that are URIs.  In theory we can use ATTRIBs for that (no Indy-Node change), but it means a bunch of agreements across Agent implementations vs. one implementation in the Indy-SDK.  Can we get a HIPE that is independent of this HIPE aimed at defining and quickly implementing that?

This is the biggest roadblock to building Agents that follow the A2A HIPEs we are defining.  Is there a way to proceed without this being in place?

2. I think this document is too loose for a HIPE.  Conceptually it makes sense, but it really needs to be tightened from speculation to firm design/specification. Do we leave this HIPE like this as an approach, but then add additional HIPEs that address the details - concepts like:

* Agent permissions (indirectly important for this but a much bigger topic),
* responsibility (is this an Indy-SDK or Agent implementation?),
* Agent Messages
* Do we need features like Before/After provability now?
* _(I'm sure there are others...)_

I do think there are potentially a number of HIPEs in here that can be implemented independently.

3. I think consideration should be given to DID Method interoperability, so there is a chance that we can use microledgers with other DID Methods. Alternatively, we could, in addition to supporting events as documented here, also support getting/putting entire current state DIDDocs.

4. It seems to me that the same Agent/Indy-SDK methods should be used for updating DIDDocs on the public ledger.  Is that planned?

5. Should a microledger implementation be at the Indy-SDK level or at the Agent level?

6. Could/should a DIDDoc be implemented (perhaps just initially) using a simpler model - a versioned/timestamped document sent from the owner to the replicas?

Not sure how to get this further, but it really feels to me like a stumbling block for Agent building progress.  I'm definitely open to ideas on how we can proceed on A2A with minimal progress on this.

Thanks!

lovesh:
@swcurran HIPE for changes in DID doc on ledger https://github.com/hyperledger/indy-hipe/pull/48

text/relationship-state-machine/README.md
```diff
@@ -0,0 +1,379 @@
...
+Implementation proposed by this HIPE will support these capabilities but cannot implement them. They will need to be part of the agent protocol, tests will need to be added in the Agent Test Suite and agent implementation will need to support them. 
+
+[event-source-descr]: https://docs.microsoft.com/en-us/azure/architecture/patterns/event-sourcing
+[merkle-tree]: https://en.wikipedia.org/wiki/Merkle_tree
+
+### Data
+The following data is expressed in the Relationship State, by extension contained in the transactions of microledger.
+
+0. DID -- The DID is immutable. Immutability is important because other layers of libindy and applications using libindy will depend on the immutability of the DID used by the party owning the DID. If a new DID is needed for a use-case, a new relationship state could be created with a new DID.
+0. Domain Endpoint -- One and only one endpoint for the domain of the owning Identity Owner. This is mutable.
```
lovesh: There can be more than 1 endpoint but the differences will just be in transport layer

text/relationship-state-machine/README.md
```diff
@@ -0,0 +1,379 @@
...
+0. Identified key -- Intended for agents but could be used for any identifiable object. Again the identifier is immutable (if a change is needed, creating a new agent is acceptable). The verkey is mutable.
+0. Authorizations -- Authorizations are attached to identifiers. These are authorizations with regards to the Relationship state and NOT for larger concerns for the application of the relationship. For example, an identifier can be authorized to onboard a new agent in the relationship state. But the relationship state cannot express permission to transfer car ownership (assuming the relationship has the type of application.)
+
+### Bootstrapping
+The relationship state must exist from the very start of a relationship. This is because data contained in the relationship state is used for communication between parties of relationship. The genesis of the relationship state (the first transaction of the microledger) MUST be communicated during the bootstrapping of the relationship.
+
+Every relationship state genesis transactions MUST ONLY consist of the following two transactions:
+
+0. DID txn
+0. Identified verkey txn
```
lovesh: This becomes part of the first (DID) txn

text/relationship-state-machine/README.md
```diff
@@ -0,0 +1,379 @@
...
+Authorizations define what changes a key can make to the relationship state. These authorizations and their corresponding rules have to be consistent between parties in the relationship. Both parties in the relationship must be able to apply the rules and always come the same conclusion. A small and simple set of authorizations and rules that govern verifiable modifications to the relationship state.
+
+Authorizations are represented as a 32-bit bit mask. This should allow for expansion of the authorization options in the future.
+
+__Permissions__:
+
+|Permission             | bitmask |
+| --------------------- |:---:|
+|ALL |                  11111111111111111111111111111111|
+|ADD_KEY |              10000000000000000000000000000000|
```
lovesh: Its efficient to start from LSB as you don't keep big numbers in memory until you need to

text/relationship-state-machine/README.md
```diff
@@ -0,0 +1,379 @@
...
+|REMOVE_KEY |           01000000000000000000000000000000|
+|UPDATE_KEY |           00100000000000000000000000000000|
+|UPDATE_AUTHORIZATION | 00010000000000000000000000000000|
```
lovesh: Should we combine `UPDATE_KEY` and `UPDATE_AUTHORIZATION` into one authorization as if you are malicious and can change someone's key then all you need to get the desired authorizations is to find a key with those authorizations and change its key to a key controlled by you. Similarly if you `UPDATE_AUTHORIZATION` but not `UPDATE_KEY`, you can change anyone's key to make it inactive, you dont need to change its key.

text/relationship-state-machine/README.md
```diff
@@ -0,0 +1,379 @@
...
+|UPDATE_ENDPOINT |      00001000000000000000000000000000|
+
+__Rules__:
+* All keys can rotate its own key
+* All keys can nullify its own key
+* ALL permission can make any modification
+* ADD_KEY permission can add a key with no authorization
+* REMOVE_KEY permission can nullify any key
+* UPDATE_AUTHORIZATION permission can modify the authorization for any key
```
lovesh: Its missing `UPDATE_KEY` from above

dhh1128:
@devin-fisher I think this has now been superseded by the peer did method spec and by the DID exchange (connection) and sync connection protocols. Do you agree, or do you want to keep it open?

devin-fisher:
Sorry, that I did not see this sooner. I agree. The useful thinking as found their way to those other efforts.

PR #35, opened 2018-08-17
merge HIPE: Cross Domain Messaging

swcurran:
Cancelled the previous PR (#21) and this is the first of three HIPEs. This one covers cross-domain messaging.


text/cross-domain-messaging/README.md
```diff
@@ -0,0 +1,329 @@
...
+    {"id": "did:sov:1234abcd#routing", "type": "RsaVerificationKey2018",  "owner": "did:sov:1234abcd","publicKeyPem": "-----BEGIN PUBLIC X…”}",
+    {"id": "did:sov:1234abcd#4", "type": "RsaVerificationKey2018",  "owner": "did:sov:1234abcd","publicKeyPem": "-----BEGIN PUBLIC 9…”}",
+    {"id": "did:sov:1234abcd#6", "type": "RsaVerificationKey2018",  "owner": "did:sov:1234abcd","publicKeyPem": "-----BEGIN PUBLIC A…”}
+  ],
+  "authentication": [
+    {"type": "RsaSignatureAuthentication2018", "publicKey": "did:sov:1234abcd#4"}
+  ],
+  "service": [
+    {"type": "Agency", "serviceEndpoint": "did:sov:fghi8377464" }
+    // or "serviceEndpoint": "https://example.com/endpoint/8377464" and add the #domain key (above)
```
kdenhartog: If we used this convention, we can have the option to add an additional parameter that would be required by the service type spec which would be something like "key" : "did:sov"fghi8377464#routing" in accordance with rule 3 of the [service endpoint section](https://w3c-ccg.github.io/did-spec/#service-endpoints) of the DID Spec. Either way works.

text/cross-domain-messaging/README.md
```diff
@@ -0,0 +1,329 @@
...
+
+### Required: The Routing Agent Processes the Outer Forward
+
+When the Routing Agent (eventually) receives the message, it determines it is the target of the outer Forward and so decrypts the `msg` value to reveal the inner "Forward" message. Recall that this inner Forward includes the full "DID#key" necessary to route the message to the intended Receiver agent. The Routing Agent uses its (implementation specific) knowledge to map from the DID#key to the Receiver, possibly via intermediary Agents. Note that the Routing Agent could itself be the Receiver and thus, have to decrypt and process the embedded message without further routing.
+
+### Required: The Receiver App Agent Decrypts/Processes the Agent Message
+
+When the intended Receiver Agent receives the message, it determines it is the target of the Forward and decrypts the payload and processes the message.
+
+### Exposed Data
```
kdenhartog: Awesome Section! In retrospect this nails my original gut feeling about how to explicitly address the information revealed (aka correlation comments) and I hadn't realized this is the perfect way to do this.

text/cross-domain-messaging/README.md
```diff
@@ -0,0 +1,329 @@
...
+
+The DIDDoc will have an additional key for each additional Agent Message Receiver.
+
+As defined in the DIDDoc Conventions HIPE, the endpoint and public key for the Domain Endpoint are found by resolving the Domain Endpoint's DID that is a service endpoint in the Receiver's DIDDoc.
+
+#### Degenerate Cases
+
+The sequence above requires there are at least three Agents within every domain, and there could be many more. However, what if there are only 2 or even 1 Agent in a domain?
+
+The HIPE requirement in those degenerate cases is that the DIDDoc still contain the same data (one endpoint, three public keys), and the domain implementation handles "Agents with multiple roles" use case. The DIDDoc could reuse the same key for different purposes, or recommended, could have an Agent with different keys for different purposes to mask the simplified Agent structure.
```
kdenhartog: Can we add something like "It is explicitly discouraged for Agencies to configure domains in this way for personal users if the single Agent is ran on an Agency's computer because this allows Agencies to impersonate users." I don't think this is something we should explicitly ban because this is very useful for enterprises. However, I would like to see us set some expectations around personal domains and this is the best place I've seen to do it so far. It may also be better to write these kind of guidelines up as a separate HIPE instead so we don't convolute this HIPE. Which do you prefer?

text/cross-domain-messaging/README.md
```diff
@@ -0,0 +1,329 @@
...
+
+Most Sender-to-Receiver messages will be sent between parties that have shared pairwise DIDs. When that is true, the Sender will (usually) AuthCrypt the message. If that is not the case, or for some other reason the Sender does not want to AuthCrypt the message, AnonCrypt will be used. In either case, the Indy-SDK `pack()` function that is used for the encryption inserts in the JOSEhdr part of its output the type of encryption used.
+
+To route the message to the Receiver, the Sender sends a 'Forward' message with the 'to' address being the DID#key of the Receiver. Also included in the message is the #key fragment of the Sender in the 'from' field. This is necessary when the message is authcrypt'd so that the receiving message can determine the Sender's key for decryption. It is assumed that the Receiver can determine the from `did` based on the `to` DID using their pairwise relationship.
```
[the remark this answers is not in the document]
dhh1128: I'm assuming that the packaging of authcrypt/anoncrypt automatically communicates the sender's key, without the agent message having to make a claim about it. Is that not the case?

text/cross-domain-messaging/README.md
```diff
@@ -0,0 +1,329 @@
...
+
+```json
+{
+  "type" : "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/routing/1.0/forward"
```
dhh1128: Is this @type?

text/cross-domain-messaging/README.md
```diff
@@ -0,0 +1,329 @@
...
+  "to"   : "did:sov:1234abcd#4"
+  "msg"  : "<pack(AgentMessage,valueOf(did:sov:1234abcd#4), privKey(A.did@A:B#1))>"
+}
```
dhh1128: Some additional and optional properties that we might consider in a forward message:
* forward-after: specify that the message should not be forwarded until the system clock shows a particular time, or until a particular interval has elapsed. This could be useful to defeat timing correlations. Particular agents might not support it.
* mix-config: artificially introduce extra hops to obscure path (use a mix network like TOR)

text/cross-domain-messaging/README.md
```diff
@@ -0,0 +1,329 @@
...
+> If the unpack() function does not support returning both the message and the Sender's public key, a "from" field in the message would need to be added so that the Sender could explicitly indicate the Sender's key within the DIDDoc.
+
+### Required: Minimize information available to the Shared Domain Endpoint
+
+We want to minimize the knowledge about the Receiver of the Agent Message for minimally trusted agents. In this case, "minimally trusted" are all agents before the designated "Routing Agent" for the Receiver. The "Routing Agent" must know the exact destination (B:did@A:B#key) of the Receiver, but Agents handling the message prior to the Routing Agent do not - they just need the DID (B:did@A:B) of the Receiver.
```
dhh1128: B.did@A:B#keyname, etc

text/cross-domain-messaging/README.md
```diff
@@ -0,0 +1,329 @@
...
+  "to"   : "did:sov:1234abcd#4"
+  "msg"  : "<pack(AgentMessage,valueOf(did:sov:1234abcd#4), privKey(A.did@A:B#1))>"
+}
+```
+
+**Notes**
+
+- the *type* value is in the precise URI format for the "forward" message type
+- Indy-SDK's "pack" function is used to AuthCrypt the message using the Receiver's public key and the Sender's private key.
+  - If the Sender knows the Receiver does not know the Sender's public key, AnonCrypt is used.
```
dhh1128: or if the sender prefers to remain anonymous

text/cross-domain-messaging/README.md
```diff
@@ -0,0 +1,329 @@
...
+  - More details about the Indy-SDK `pack()` function can be found in the `Wire Messages` HIPE (*reference to be provided*).
+- The Indy-SDK's "unpack()" function **must** return the public key of the private key used to sign the message. See the note below for the background on this.
```
dhh1128: Yes, this is the way authdecrypt works today.

text/cross-domain-messaging/README.md
```diff
@@ -0,0 +1,329 @@
...
+
+> The bullet above about the unpack() function returning the signer's public key deserves some additional attention. The Receiver of the message knows form the "to" field the DID to which the message was sent. From that, the Receiver is expected to be able to determine the DID of the Sender, and from that, access the Sender's DIDDoc. However, knowing the DIDDoc is not enough to know from whom the message was sent - which key was used to send the message, and hence, which Agent controls the Sending private key. This is information must be known to the Receiver so that they know which key to use in responding to the arriving Message.
```
dhh1128: I don't think that knowing the specific sender will tell the receiver which key to use in responding to the arriving message; not every message emitted by 1 should be replied to at 1, or even replied to at all. I think "reply-to" is an optional message decorator that could be added.

text/cross-domain-messaging/README.md
```diff
@@ -0,0 +1,329 @@
...
+> If the unpack() function does not support returning both the message and the Sender's public key, a "from" field in the message would need to be added so that the Sender could explicitly indicate the Sender's key within the DIDDoc.
```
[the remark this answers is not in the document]
dhh1128: The sending key has to be returned by authdecrypt or by headers that authdecrypt uses; it can't be in the agent-level message.

text/cross-domain-messaging/README.md
```diff
@@ -0,0 +1,329 @@
...
+  "type" : "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/routing/1.0/forward"
+  "to"   : "did:sov:1234abcd"
+  "msg"  : "<pack(ForwardMessage,valueOf(did:sov:1234abcd#routing))>"
+}
+```
+
+**Notes**
+
+- Indy-SDK's "`pack()`" function is used to Anon encrypt the message using the Routing Agent's public key.
+  - There is no need for the unpack() function to return the Sender's public key, since the Routing Agent does not need to know that information.
```
dhh1128: Agreed. Note that this implies that it is impossible for an agency to do spam filtering on messages. I am okay with this, but just wanted to point it out.

text/cross-domain-messaging/README.md
```diff
@@ -0,0 +1,329 @@
...
+
+The Sender can now send the message on its way via the first of the Wire messages. In our example, it sends the message to 2, who in turn sends it to 8. That of course, is arbitrary - the Sender's Domain could have any configuration of Agents.
```
dhh1128: Your use of "it" here could benefit from clarification. The sender sends the outer forward message (and all the inner stuff it contains) to 2. 2 peels back 1 layer of the envelope and sends the payload from its forward message on to the next hop, etc.

text/cross-domain-messaging/README.md
```diff
@@ -0,0 +1,329 @@
...
+
+### Required: The Routing Agent Processes the Outer Forward
+
+When the Routing Agent (eventually) receives the message, it determines it is the target of the outer Forward and so decrypts the `msg` value to reveal the inner "Forward" message. Recall that this inner Forward includes the full "DID#key" necessary to route the message to the intended Receiver agent. The Routing Agent uses its (implementation specific) knowledge to map from the DID#key to the Receiver, possibly via intermediary Agents. Note that the Routing Agent could itself be the Receiver and thus, have to decrypt and process the embedded message without further routing.
```
dhh1128: Now I feel puzzled. The agency endpoint receives the outer forward message, anondecrypts it, and sends the inner forward message to the routing agent. So the sentence that says "it determines it is the target of the outer forward" doesn't make sense. The routing agent shouldn't see the outer forward.

text/cross-domain-messaging/README.md
```diff
@@ -0,0 +1,329 @@
...
+
+### Required: The Receiver App Agent Decrypts/Processes the Agent Message
+
+When the intended Receiver Agent receives the message, it determines it is the target of the Forward and decrypts the payload and processes the message.
```
dhh1128: Again, the receiver doesn't receive a forward; it receives what's inside the forward.

text/cross-domain-messaging/README.md
```diff
@@ -0,0 +1,329 @@
...
+  - The Domain Endpoint's physical endpoint and public key
+  - The Routing Agent's public key
+  - The Receiver of the Agent Message's public key
+
+Thus, every Messaging DIDDoc is assumed to have at least one endpoint and three keys:
+
+- The endpoint for the Domain Endpoint
+- The public key for the Domain Endpoint
+- The public key for the Routing Agent
+- The public key for the Receiver Agent
```
dhh1128: Are we explicitly saying that a routing agent's key cannot be used to receive as well? If yes, then 3 keys would be required. If no, then 2 keys would still be a valid scenario (but would only support routing-agent-as receiver).

text/cross-domain-messaging/README.md
```diff
@@ -0,0 +1,329 @@
...
+
+## Core:Routing:1.0:Forward
+
+The core message type "forward", version 1.0 of the "routing" family is defined in this HIPE. An example of the message is the following:
+
+```json
+{
+  "type" : "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/routing/1.0/forward"
+  "to"   : "did:sov:1234abcd#4"
+  "from" : "#1"  // Optional: The key used to AuthCrypt the message
```
dhh1128: I don't want this field to even be allowed if its definition is "the key used to authcrypt". That shouldn't be a concern of an application message; it should be handled by packaging. However, I'd be happy to have this field if the intent of it were to disclose the DID of the sender to an intermediate router (e.g., so a routing agent could apply policy tailored to particular senders).

text/cross-domain-messaging/README.md
```diff
@@ -0,0 +1,329 @@
...
+The `to` field is required and takes one of two forms:
+
+- A DID without a key reference (e.g. `did:sov:1234abcd`)
+- A DID with a key reference (e.g. `did:sov:1234abcd#4`)
+
+The first form is used when sending forward messages across one or more agents that do not need to know the details of a domain. The Receiver of the message is the designated `Routing Agent` in the Receiver Domain, as it controls the key used to decrypt messages sent to the domain, but not to a specific Agent.
+
+The second form is used when the precise key (and hence, the Agent controlling that key) is used to encrypt the Agent Message placed in the `msg` field.
+
+The `from` field is used and required only when a message is AuthCrypt'd by the Sender. The Receiver of the message can determine, based on the DID in the `to` field the corresponding pairwise DID of the Sender. However, it cannot from that determine the specific key used by the Sender to AuthCrypt the message. The `from` field provides that missing information. The exclusion of the DID in the from field reduces the information visible in the message.
```
dhh1128: see my note above; I don't want this

text/cross-domain-messaging/README.md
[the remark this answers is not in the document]
swcurran: There is debate in the W3C DID group about whether there should be a standard way to link between a service endpoint and a key.  I think we should push hard for that.

text/cross-domain-messaging/README.md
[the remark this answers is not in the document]
swcurran: Updated to reflect that the authcrypt version of the unpack() function will return the message and the senders public key.

text/cross-domain-messaging/README.md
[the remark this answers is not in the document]
swcurran: Updated to reflect Daniel's comment.

text/cross-domain-messaging/README.md
swcurran: Updated the phrase to use SHOULD and not mention the option of not doing that.

text/cross-domain-messaging/README.md
```diff
@@ -0,0 +1,329 @@
...
+
+The goal of this HIPE is to define the rules that domains must adhere to enable the delivery of Agent messages from a Sending Agent to a Receiver Agent in a secure and privacy-preserving manner.
+
+# Motivation
+[motivation]: #motivation
+
+The purpose of this HIPE and it's related HIPEs is to define a layered Messaging protocol such that we can ignore the transport of messages as we discuss the much richer Agent Messaging types and interactions. That is, we can assume that there is no need to include in an Agent message anything about how to route the message to the Receiver - it just magically happens. Alice (via her App Agent) sends a message to Bob, and (because of implementations based on this series of HIPEs) we can ignore how the actual message got to Bob's App Agent.
```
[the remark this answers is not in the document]
swcurran: Sigh

text/cross-domain-messaging/README.md
```diff
@@ -0,0 +1,329 @@
...
+The following are assumptions upon which this HIPE is predicated.
+
+### Terminology
+
+The following terms are used in this HIPE with the following meanings:
+
+- Domain - a set of Agents collaborating on behalf of an Identity
+  - It's assumed that the Agents of a Domain were developed by a single vendor and so may have implementation-specific mechanisms for tracking extra information one another.
+  - An example of two Domains is provided in the image below.
+- App Agents - the Agents involved in sending (creating) and receiving (processing) an Agent Message
```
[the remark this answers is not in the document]
swcurran: I've changed them all to "Edge Agent".  Still don't like it, but they are consistent.

text/cross-domain-messaging/README.md
```diff
@@ -0,0 +1,329 @@
...
+  - Sender - the Agent sending an Agent Message
+  - Receiver - the Agent receiving an Agent Message
```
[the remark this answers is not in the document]
swcurran: Leaving it for the reason stated.

text/cross-domain-messaging/README.md
```diff
@@ -0,0 +1,329 @@
...
+  - **Note**: A message may pass through many Agents between the Sender and Receiver
+- Domain Endpoint - a shared physical endpoint for messages into domains
```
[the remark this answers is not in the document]
swcurran: I like that.

text/cross-domain-messaging/README.md
```diff
@@ -0,0 +1,329 @@
...
+  - Shared by many identities (e.g. https://endpoint.agentsRus.com)
+  - Agency - the handler for messages sent to the Domain Endpoint.
+- Routing Agent - the single identity-controlled entry point for a Domain per relationship
+  - A message delivered to a Domain Endpoint is **always** passed directly to the Routing Agent for a Domain
```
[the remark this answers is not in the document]
swcurran: Reworded to clarify.

text/cross-domain-messaging/README.md
```diff
@@ -0,0 +1,329 @@
...
+- DID - reference to the literal Decentralized ID text
+  - e.g. did:sov:1234abcd
+- DID#key - reference to the DID appended with "#" and a specific key from the DIDDoc
```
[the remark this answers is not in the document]
swcurran: Updated

text/cross-domain-messaging/README.md
```diff
@@ -0,0 +1,329 @@
...
+}
+```
+
+For the purposes of this discussion we are defining the message flow to be:
+
+> 1 --> 2 --> 8 --> 9 --> 3 --> 4
+
+However, that flow is arbitrary. Even so, some Wire Message hops are required:
+
+- 1 is the Sender and so must send the first message.
```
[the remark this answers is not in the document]
swcurran: Updated

text/cross-domain-messaging/README.md
```diff
@@ -0,0 +1,329 @@
...
+
+## Cross Domain Interoperability
+
+A key goal for interoperability is that we want other domains to know just enough about the configuration of a domain to which they are delivering a message, but no more. The following walks through those minimum requirements.
+
+### Required: The DID and DIDDoc
+
+As noted above, the Sender of an Agent to Agent Message has the DID of the Receiver, and knows the key from the DIDDoc to use for the Receiver's Agents.
+
+> Example: Alice wants to send a message from her phone (1) to Bob's phone (4). She has Bob's B:did@A:B, the DID/DIDDoc Bob created and gave to Alice to use for their relationship. Alice created A:did@A:B and gave that to Bob, but we don't need to use that in this example. The contents of the DIDDoc for B:did@A:B are presented above.
```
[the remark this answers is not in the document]
swcurran: That's what I used.  Did you mean to say something different?

text/cross-domain-messaging/README.md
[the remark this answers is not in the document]
swcurran: Fixed

text/cross-domain-messaging/README.md
swcurran: Added in. I'm still not quite sure when there should be a literal @ in the field or not.  I'm further confused by our use of @ in decorators - eg. @thread. That seems to be an overloading of the symbol.  But as I say, I don't quite get the rules yet.

swcurran: About the "from" - I had removed it from the JSON and explained it below, but missed the reference above the JSON. Removed.

text/cross-domain-messaging/README.md
swcurran: Left off of the spec for routing/1.0/forward.  Added as a "to be considered" in the Rationale and Alternatives section.

text/cross-domain-messaging/README.md
swcurran: Fixed

text/cross-domain-messaging/README.md
swcurran: Added

text/cross-domain-messaging/README.md
swcurran: With AnonCrypt, I don't think the Sender's public key is known, is it?  Checking the AonoCrypt definition, that is the case.

text/cross-domain-messaging/README.md
swcurran: No. The Domain Endpoint receives the Wire Message that wraps the Outer Forward.  Thus, when the unpack() of the Wire Message is complete, the Domain Endpoint sees the outer forward ("to": "did") and knows to send it to the Routing Agent.  Only the Routing Agent can decrypt the "msg", which contains the Inner Forward.

text/cross-domain-messaging/README.md
swcurran: I disagree. The Receiver App gets all three fields of the forward and decrypts the message. In theory you could change the message type to be (I don't know...) "message", to indicate it is not a forward, but you are still sending the "type" and "msg" fields.  By not sending the "to" field (not a significant savings), the Receiver loses an easy way to determine the Sender (using the pairwise relationship).

text/cross-domain-messaging/README.md
swcurran: Per the degenerate case described in this and the DIDDoc HIPE, even if the Routing Agent is also the Receiver Agent, the DIDDoc MUST be configured to have all three and SHOULD be configured to use different keys for all roles.

text/cross-domain-messaging/README.md
swcurran: Inadvertently left in during the last round of updates.  Removed.

text/cross-domain-messaging/README.md
swcurran: Removed.

swcurran:
Updated the HIPE to address the comments and adjusted as necessary any "MUST" and "SHOULD" references.

text/cross-domain-messaging/README.md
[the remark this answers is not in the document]
kdenhartog: This is true in the case of how we're currently doing this, but with JWM stuff there will be a header that identifies the sender's key. I was planning to have the unpack() function extracting that data and then passing it into auth_decrypt, but what you're describing may be the better approach in order to reduce complexity. Also, how would a receiver know who to send back to if the message is anoncrypt()/anon_decrypt rather than auth? Would it have to be contained in the message?

text/cross-domain-messaging/README.md
kdenhartog: Sam and I were going over this stuff yesterday. JWMs contain all the information in Forwards right now, so I'm playing with the idea that JWMs would replace the forward message and what that would look like. This is one of the efficiency things I mentioned on the call today.

text/cross-domain-messaging/README.md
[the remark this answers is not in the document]
kdenhartog: I agree, this would be a good topic for conversation at RWoT. I haven't heard CCG talk about DID spec in some time, so I stopped showing up to the CCG. If it's added to the agenda I'll join and discuss.

text/cross-domain-messaging/README.md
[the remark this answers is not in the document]
swcurran: Agree that if AnonCrypt is used, the Receiver will not know who actually sent the message without a "from" or having the specific message notify the user.  I don't like relying on having a from field on every message type - it feels like it should be part of the envelope.

kdenhartog:
Can we move this to FCP? The JSON structure of this (Forward message) hasn't change any at this point. We can go back and change other stuff once it's accepted so we can start implementing this stuff.

PR #36, opened 2018-08-17
propose HIPE: Wire messages

swcurran:
Cancelled the previous PR (#21) and this is the second of three HIPEs. This one covers wire messages.


text/wire-messages/README.md
```diff
@@ -23,13 +23,10 @@ The purpose of this HIPE is to define how an Agent that needs to transport an ar
 
 For the purposes of this HIPE, the following are assumed about the sending and delivery of Wire Messages.
 
-- Each Agent sending a Wire Message knows to what Agent the message is to be sent.
-
-- Each Agent knows what encryption (if any) is appropriate for the message.
-
+- Each Agent sending a Wire Message knows to what Agent the wire message is to be sent.
+- Each Agent knows what encryption (if any) is appropriate for the wire message.
```
kdenhartog: I'm a little uncomfortable with this assumption right now. How can we be certain that agents will always know the type of encryption used? I was under the assumption that the JOSE header would list out the type of encryption similar to how JWEs do this, but I don't think it's necessary if we know a way to do without it.

text/wire-messages/README.md
```diff
@@ -0,0 +1,209 @@
...
+
+# Tutorial
+[tutorial]: #tutorial
+
+## Assumptions
+
+For the purposes of this HIPE, the following are assumed about the sending and delivery of Wire Messages.
+
+- Each Agent sending a Wire Message knows to what Agent the wire message is to be sent.
+- Each Agent knows what encryption (if any) is appropriate for the wire message.
```
kdenhartog: I'm a little uncomfortable with this assumption right now. How can we be certain that agents will always know the type of encryption used? I was under the assumption that the JOSE header would list out the type of encryption similar to how JWEs do this, but I don't think it's necessary if we know a way to do without it.

text/wire-messages/README.md
```diff
@@ -0,0 +1,209 @@
...
+// Receiving
+tmsg = recv(myEndpoint)
+msg = unpack(tmsg, myPrivKey)
+```
+
+Once the `msg` value has been extracted from the Wire Message, it is processed as an Agent Message. That is, its type is evaluated and the entire message given to a handler configured for that message type.
+
+### The pack()/unpack() Functions
+
+The pack() function is implemented in the Indy-SDK and will evolve over time. The initial instance of pack() includes three variations, only the first two of which are used for Wire Messages. The third is used by the Routing family of Agent Messages (defined in the Cross Domain Messaging HIPE).
```
kdenhartog: I don't believe this is the way libvcx has implemented this, so making a note here that will have to change libvcx's implementation to support this. In the future, Compact JWMs should look similar (stuff packed together right now will be separated and period seperated) to this as well, so I think refactoring the libvcx is a good approach in the interim.

text/wire-messages/README.md
```diff
@@ -0,0 +1,209 @@
...
+  - 3 App Agents - "4", "5" and "6"
+    - "6" is an App Agent in the cloud, "4" and "5" are physical devices.
+  - 1 Routing Agent - "3"
+  - 1 Domain Endpoint - "9"
+
+### Bob's DID for his Relationship with Alice
+
+Bob’s domain has 3 devices he uses for processing messages - two phones (4 and 5) and a cloud-based agent (6). However, in Bob's relationship with Alice, he ONLY uses one phone (4) and the cloud-based agent (6). Thus the key for device 5 is left out of the DIDDoc (see below).
+
+Note that the key for the Routing Agent (3) is called "routing". This is an example of the kind of convention needed to allow the Sender's agents to know the keys for Agents with a designated role in the receiving domain.
```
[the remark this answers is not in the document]
kdenhartog: Noting this question is addressed within the [Cross Domain Messaging HIPE](https://github.com/hyperledger/indy-hipe/blob/2b15fe24f1e97bd0a1b7e1d4cc31ae39d20cc3b7/text/cross-domain-messaging/README.md#required-cross-domain-wire-message-encryption). Will change to approved review.

text/wire-messages/README.md
```diff
@@ -0,0 +1,209 @@
...
+
+This HIPE addresses only Wire Messages. There are a series of related HIPEs that combine with this HIPE to address the interoperability, including in particular, Agent Messages, DIDDoc Conventions, and Cross Domain Messaging. Those HIPEs should be considered together in understanding Agent-to-Agent interoperability.
+
+# Motivation
+[motivation]: #motivation
+
+The purpose of this HIPE is to define how an Agent that needs to transport an arbitrary Agent Message delivers it to another Agent through a direct (point-to-point) communication channel. A message created by a Sender Agent and intended for a Receiver Agent will usually be sent multiple times between Agents via Wire Messages in order to reach it's ultimate destination.
```
dhh1128: "it's ultimate destination" --> "its ultimate destination"

text/wire-messages/README.md
```diff
@@ -0,0 +1,209 @@
...
+[tutorial]: #tutorial
+
+## Assumptions
+
+For the purposes of this HIPE, the following are assumed about the sending and delivery of Wire Messages.
+
+- Each Agent sending a Wire Message knows to what Agent the wire message is to be sent.
+- Each Agent knows what encryption (if any) is appropriate for the wire message.
+- If encryption is to be used, the sending Agent knows the appropriate public key (of a keypair controlled by the receiving Agent) to use.
+- The sending Agent knows the physical endpoint to use for the receiver, and the appropriate Internet Transport Protocol (https, zmq, etc.) to use in delivering the wire message.
```
dhh1128: remove "Internet" -- some protocols don't use the internet

text/wire-messages/README.md
```diff
@@ -0,0 +1,209 @@
...
+
+> The term "shared Domain Endpoint" is defined in the Cross Domain Messaging HIPE. In short, this is assumed to be an Agency endpoint that receives messages for many Identities, each with many DIDs.
```
dhh1128: see my note in the other HIPE about being too strong with "shared" in "shared Domain Endpoint"

text/wire-messages/README.md
```diff
@@ -0,0 +1,209 @@
...
+## Example: Domain and DIDDoc
+
+The example of Alice and Bob's domains are used for illustrative purposes in defining this HIPE.
+
+![Example Domains: Alice and Bob](domains.jpg)
+
+In the diagram above:
+
+- Alice has
+  - 1 App Agent - "1"
```
dhh1128: App Agent was not defined in this HIPE or in the one about DID Docs.

text/wire-messages/README.md
```diff
@@ -0,0 +1,209 @@
...
+}
+```
+
+For the purposes of this discussion we are defining the Wire Message Agent message flow to be:
+
+> 1 --> 2 --> 8 --> 9 --> 3 --> 4
+
+However, that flow is arbitrary. Even so, some Wire Message hops are required:
+
+- 1 is the Sender and so must send the first message.
```
dhh1128: Not all sending from Alice needs to start with 1. So 1 is arbitrarily chosen, not required. (However, 1 will certainly be the most common sender.)

text/wire-messages/README.md
```diff
@@ -0,0 +1,209 @@
...
+- 9 is the Domain Endpoint of Bob's domain and so must receive the message
+- 4 is the Receiver and so must receive the message.
```
dhh1128: Any of Bob's agents (including the routing agent) could be the receiver(s). So again, 4 is arbitrarily chose. The only thing that's truly not arbitrary is the use of 9.

text/wire-messages/README.md
dhh1128: Ah, I see where it was defined in your cross-domain HIPE.

text/wire-messages/README.md
swcurran: Yes.  I think a JIRA is needed for this.  Since the third form requires two values be returned, I would recommend that the return value be JSON in all cases, to be consistent.

text/wire-messages/README.md
swcurran: Begrudgingly changed to "Edge Agent" :-)

text/wire-messages/README.md
swcurran: Clarified

text/wire-messages/README.md
swcurran: Clarified

swcurran:
Updated HIPE per comments and checked use of MUST/SHOULD, etc.

text/wire-messages/README.md
swcurran: Not in these cases. The use cases don't have an opinion on the Trust of the Agents, or even why the Cloud Edge Agent exists.  Trust level may be a concept we get to, but so far we've not yet come across a concrete reason for it.

dhh1128:
Just a comment for those who are curious: here's info about how this PR relates to PR #43 : https://github.com/hyperledger/indy-hipe/pull/43#issuecomment-436075445

kdenhartog:
My intention for the AMES HIPE #43 is to replace this HIPE. I've included the information I thought would be relevant, but if there's more information that I should include can you let me know? If there's no additional information I should include then I'd propose closing this PR.

swcurran:
Agree that this one should be closed.  It's work is done...rip.

PR #38, opened 2018-09-05
Add Forward Secrecy to Agent-to-Agent communication

mikelodder7:


dbluhm:
[rendered text](https://github.com/mikelodder7/indy-hipe/tree/master/text/a2a-forward-secrecy)

vimmerru:
I have few questions:

1. Edge-2-Edge pairwise communication is only part of communication types supported by Indy. We have also at least:
 - Public DIDs published on the ledger with DIDDoc
 - Communication with agencies
 - Communication between edge agent and supplementary (cloud) agents 
 - Communication between (cloud) agents 

Supporting of Signal between each chain link seems overkill and doesn't possible for some links. Do you think we need to make Signal optional or use Signal only for Edge-2-Edge messages?

2. Can we just use continuous key rotation in microledgers to achieve some kind of forward security instead?

vimmerru:
@mikelodder7 Could you also clarify how this approach can be integrated with DIDDoc spec and agent domains? As i understand existing Signal implementation binds ratchet to one edge-device and it is impossible to receive messages on different edge-devices.

mikelodder7:
I’m intending this to be between edge agents whether in domain or out of domain.

I don’t want to have to rotate the micro ledger keys after every message. The transactions for the micro ledger would become very big if we did that and would bloat messaging as with each message you send you would also have to send the signed transaction to enforce the new key

mikelodder7:
This protocol would not be used to talk to the ledger. That would be overkill and the ledger would have to maintain state for it which we don’t need to do.

To talk to agencies and the ledger which are online, I envision we can just use TLS with forward secret ciphers. So cloud agent 1 can use TLS to send a message to cloud agent 2. Whenever we have synchronous messaging it makes sense to just use TLS

mikelodder7:
@vimmerru as I understand it, we already have to maintain keys for each edge agent for every relationship. This would just employ the signal protocol also.

mikelodder7:
You raise good points so I will add clarity text to reflect it

mikelodder7:
Added sections to reflect @vimmerru's questions

TelegramSam:
"The microledger maintains transactions forever" -- Does it though? I think microledgers can agree to choose a newer message as the new genesis transaction, and it could prune all the history?

TelegramSam:
Also, I'm curious if message by message rotation every single time is what provides the forward secrecy, or if 'frequent' rotation is what provides it. Can you elaborate on that?

mikelodder7:
We haven't defined when micro ledger transactions should be deleted so until we do I assume all transaction from the dawn of time. I'm hoping we could prune transactions but nevertheless we don't want rotating the keys for the micro ledger with every transaction because that is essentially recreating the Signal protocol but Signal has been vetted and this hasn't. I also believe it wouldn't be as secure as Signal to rotate with every transaction because we are signing a transaction, encrypting it, hoping the previous key still works and hasn't been subject to man-in-the-middle. Signal is resistant to man-in-the-middle and only vulnerable during channel establishment.

mikelodder7:
Forward secrecy is provided by using the message key once. If the long term identity keys are compromised then one cannot derive the session keys and decrypt the messages. Even if they discover a single message key, its only helpful for that message and not going **forward**.

PR #45, opened 2018-09-27
FCP INDY-1704: create a multi-sig-actions HIPE doc.

sergey-shilov:
Signed-off-by: Sergey Shilov <sergey.shilov@dsr-company.com>


text/multi-sig-trustee-actions/README.md
```diff
@@ -0,0 +1,139 @@
+- Name: multi-sig-trustee-actions
+- Author: Sergey Shilov
+- Start Date: 2018-09-27
+- PR:
+- Jira Issue:
```
ashcherbakov: Please create and specify the Epic here

text/multi-sig-trustee-actions/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+# Summary
+[summary]: #summary
+
+The indy-node should require several signatures of Trustees for the
```
ashcherbakov: I would make the Hipe more generic, and say about actions requiring multiple signers:

There are  important and potentially dangerous actions in Indy which require a quorum of signers (Trustees, Stewards, etc.).
The purpose of this PR is to describe approach on how multi-signed actions are going to be supported in Indy.

text/multi-sig-trustee-actions/README.md
```diff
@@ -0,0 +1,139 @@
...
+same transaction.
+
+Current implementation of libindy has an API call to add a signature
+for the transaction. When such multi-signed transaction is received
+by the indy-node then all signatures of this transaction are verified.
+But preparation of multi-signed transaction is up to application.
+
+Also it is required to use the same source DID to have ability to
+uniquely identify this transaction in case of sending to the different
+nodes by different clients.
```
ashcherbakov: Please describe steps each Client (Trustee) need to do to send the transaction (including what SDK methods need to be called), plus what actions Ledger (each Node) does to verify the multi-signature. Example:
1) Trustee1 creates a transaction payload and calls metjhod_x to sign it. The result is a JSON containing a signature. Then Trustee1 transfers the JSON to Trustee2.
2) trustee 2 checks (optionally) a signature of Trustee1, and sign it calling method_X. The result is a JSON containing two signatures. Then Trustee2 transfers the JSON to Trustee3
.....
3) Each Node gets the txn, ans as part of static validation (before ordering and consensus steps) verifies the multi-sigs as follows:
- check what action is it
- gets how many signatures are required for this action
- verifies each signature
- check that the numbr of signatures is not less that the required threshold

text/multi-sig-trustee-actions/README.md
```diff
@@ -0,0 +1,139 @@
...
+I propose to divide the implementation into 2 phases:
+
+ - Phase 1: extend the static *auth_map* structure (short term solution)
+ - Phase 2: move the *auth_map* structure to config ledger (long term solution)
+
+## Phase 1
+
+It would be more efficient for whole stack to implement a solution with
+static rules first. The plan for this phase is:
+1. Modify and extend the *auth_map* structure to store the authorisation
```
ashcherbakov: Please mention a task to support %-based strategy (the same as in PoA)

text/multi-sig-trustee-actions/README.md
```diff
@@ -0,0 +1,139 @@
...
+[alternatives]: #alternatives
+
+Instead of multi-signed transaction it is possible to require several
+single-signed transactions from trustees. But:
+ - this requires to store them on the node side
+ - this requires to maintain some state of transaction
+ - this does not correspond to current protocol
+ - this requires a lot of changes of the indy-node codebase
+
+So this alternative is much worse than proposed solution.
```
ashcherbakov: I don;t think it's worse. It's actually better (so called Voting Protocol). It's just to complex and hard to be implemented, and we think that the proposed solution should work well in a short-term.

devin-fisher:
Seems that the weak link is the config txn that configures this policy. Some description of how this will work is important. I assume that only Trustees can submit config TXNs. Is that right? Secondly, how many Trustees will be required to change the config TXN? 

If the policy is that it takes three trustees to add a trustee but it only takes one trustee to change the number of trustees it takes to add a trustee, then, in fact, it only takes one trustee to add a trustee.

devin-fisher:
It would seem that adding trustee would always have to require the most signers, otherwise, all other actions reduce to this number. For example, if I wanted to upgrade the ledger to a new version (a bad version with a backdoor). But I only have two trustees colluding with me and the update txn requires three trustees. But adding a trustee requires only two. I could quickly add a new trustee and then I would have the required three and I could then issue the upgrade txn.

text/multi-sig-trustee-actions/README.md
sergey-shilov: Done

text/multi-sig-trustee-actions/README.md
sergey-shilov: Done

text/multi-sig-trustee-actions/README.md
sergey-shilov: Done

text/multi-sig-trustee-actions/README.md
sergey-shilov: Done

sergey-shilov:
> Seems that the weak link is the config txn that configures this policy. Some description of how this will work is important. I assume that only Trustees can submit config TXNs. Is that right? Secondly, how many Trustees will be required to change the config TXN?
> If the policy is that it takes three trustees to add a trustee but it only takes one trustee to change the number of trustees it takes to add a trustee, then, in fact, it only takes one trustee to add a trustee.

It's a question of genesis transactions for config ledger. Concrete values of quorums is a question for discussion and agreement of stakeholders. We can define the lowest thresholds as a constants in the indy-node code base, but I don't think that it is right solution as we want to have it configurable.

sergey-shilov:
> It would seem that adding trustee would always have to require the most signers, otherwise, all other actions reduce to this number.

Yes, sure.

TelegramSam:
Is the method of transferring between trustees out of scope for the doc? 
`Then Trustee1 transfers the JSON to Trustee2` appears in muliple places, but that transmission method is never discussed.

ashcherbakov:
@swcurran 
I believe exact steps (and calls) for libindy are provided in Tutorial section: https://github.com/hyperledger/indy-hipe/blob/10fc1f46631de4c86c95717cfa15446f67f42646/text/multi-sig-actions/README.md#tutorial

Is it enough?

ashcherbakov:
@devin-fisher @sergey-shilov 
> I assume that only Trustees can submit config TXNs. Is that right? 
I think yes.
> Secondly, how many Trustees will be required to change the config TXN?
I believe we can use other strategies (hard-coded in config percent-based or constant-based number of Trustees required) for the first txn to config ledger. After this we can use the current config txn value for the update.

text/multi-sig-actions/README.md
```diff
@@ -0,0 +1,169 @@
...
+same source DID (action initiator) and signed by required number of
+trustees**. Otherwise the transaction will be rejected. It means that
+the client application should take care about multiple signing of the
+same transaction.
+
+Current implementation of libindy has an API call
+*indy_multi_sign_request()* to add a signature for the transaction.
+When such multi-signed transaction is received by the indy-node then
+all signatures of this transaction are verified. But preparation of
+multi-signed transaction is up to application.
```
kdenhartog: I take this to mean, the SDK won't handle the passing of the partially signed messages, but rather should be handled by an application (agent). Am I interpreting this correctly?

text/multi-sig-actions/README.md
```diff
@@ -0,0 +1,169 @@
...
+ - find out the structure of "actions" and their definition
+ - apply them for incoming transactions as a part of authorisation
+ - store them in config ledger to make them configurable
+
+For now we have a static structure called *auth_map*. This structure
+defines permissions for various actions. So we can use this data
+structure as a base for new rules definition with some changes and
+extensions.
+
+I propose to divide the implementation into 2 phases:
```
kdenhartog: Will this two phase implementation create a breaking change? It doesn't seem like it would, but I'm not familiar enough with the indy-node codebase to know whether this would.

PR #46, opened 2018-10-12
propose HIPE: Agent connection protocol

mikelodder7:
This is meant to replace [HIPE 18](https://github.com/hyperledger/indy-hipe/pull/18)

[Rendered Text](https://github.com/mikelodder7/indy-hipe/blob/agent-connection-protocol/text/agent-connection-protocol/README.md)


mikelodder7:
[Rendered Text](https://github.com/mikelodder7/indy-hipe/tree/agent-connection-protocol/text/agent-connection-protocol)

dbluhm:
It seems as though it might make sense to leave "new" off the front of each of the message types. It feels a little redundant but I don't feel strongly about this if you think it should stay that way.

mikelodder7:
> It seems as though it might make sense to leave "new" off the front of each of the message types. It feels a little redundant but I don't feel strongly about this if you think it should stay that way.

Done

dbluhm:
Another minor comment regarding the types: I think we might want to make `connection` the message family and move that from off the front of the type.

So, this `"@type": "did:sov:1234567890;spec/messagefamily/1.0/connectionofferunencrypted"` would become `"@type": "did:sov:1234567890;spec/connection/1.0/offerunencrypted"`.

swcurran:
Since the intention of the connection request (and response) is to establish a pairwise relationship with the other party, I think the exchanged data for these messages should not be a public key but rather a DID and DIDDoc. I would think the other party would also need to know what key entry it should be using to respond, so as that the response would be sent to the correct agent. The connection being established is (I believe) between two domains, not just two agents (based on the requirements in the Cross-Domain and DIDDoc Convention HIPEs) and so this protocol should account for that.

In saying that, I'm assuming that the "message" is just a piece of verification text being exchanged - correct?

That implies this HIPE blurs into the Relationship State Machine HIPE/protocol. My preference is that the RSM is an internal implementation (not exposed at the cross-domain level) and so other than knowing that somehow the set of agents that make up the domain have to manage the RSM, it has little impact on this HIPE - other than the requirement that the implementation provide a DID and DIDDoc for these messages to send.

swcurran:
We've come across a couple of use cases recently where an introduction mechanism might be useful from a user experience perspective. An "introduction" in this case would be where two identities have a relationship, and one (Alice) wants to introduce the other (Bob) to a third party (Carol). Alice and Bob, and Alice and Carol already have relationships and the goal is for Bob and Carol to have a relationship (that doesn't include Alice). It seems that the use of unencrypted communications and public DIDs might be able to be avoided if a temporary public key can flow from Bob through Alice to Carol in place of the "out of band connection". I think it might make sense to standardize, could reduce correlation opportunities and could improve user experience. Risk is that it be able to be abused, so that needs to be considered.

Has this been considered? Good/bad idea?

mikelodder7:
So this protocol uses “temporary” keys. The public key here is an ephemeral key so not correlation is created. It also adds forward secrecy. This does not handle routing I’m assuming that is handled at a lower level. By doing introductions this way you either have 0-RTT with public DIDs and 1-RTT otherwise.

dbluhm:
@swcurran I added some details which might address your concerns. What this protocol is really trying to establish is not *what* needs to be sent to establish the pairwise relationship but *how* that information is sent. In place of the DID and Verification key in the connection message, we could just as easily put just a DID to be looked up on the ledger or a DIDDoc.

Our focus at the moment is to establish connections with what we have today (for a functional demo at IIW) which, unfortunately, means that we won't be sending a full DIDDoc because we don't have DIDDoc support in Indy-sdk yet. Additionally, before introducing the complexities of routing through domains, we want to demonstrate the basics of messaging. As we see things like Microledger and DIDDoc support hit the SDK, I think we can start to have more HIPEs addressing exactly what content is sent as part of a connection establishment.

By keeping the scope of this HIPE narrow, we can focus on the mechanics and crypto of the exchange itself.

dbluhm:
Feel free to correct me if I'm off, @mikelodder7

swcurran:
Hey @dbluhm, reasonable I guess, but I don't think HIPEs should be written to deal with short term deliverables.  I think it's great that your team is working on an immediate deliverable (can't wait to see it!), but I don't think a HIPE should be constrained by that deliverable.  HIPEs (and **especially** this HIPE IMHO) should be driving what core indy-sdk functionality needs to be implemented, so work can be prioritized on that.  HIPEs shouldn't propose unrealistic "cool, but ain't never gonna happen" approaches, but they shouldn't be limited to what can be done today.

I'd prefer to see a complete HIPE for Establishing Connections, with perhaps a "Current Status" section that deals with what can be implemented in the short term.  Ideally, there should also be a "Dependent On" section that points to other HIPEs/features needed.

It's not clear my suggestion is right (although it is how I understand the HIPE process).  I think if we do go with the more pragmatic approach you are proposing, then we should back off on several of the other HIPEs that have been proposed - especially the Cross-Domain and DIDDoc Conventions ones, as they are contradictory to this HIPE - and this one has code.

Finally - I would add that this to me shows the urgency of getting DIDDoc support in at least indy-sdk and likely indy-node.

mikelodder7:
> Hey @dbluhm, reasonable I guess, but I don't think HIPEs should be written to deal with short term deliverables. I think it's great that your team is working on an immediate deliverable (can't wait to see it!), but I don't think a HIPE should be constrained by that deliverable. HIPEs (and **especially** this HIPE IMHO) should be driving what core indy-sdk functionality needs to be implemented, so work can be prioritized on that. HIPEs shouldn't propose unrealistic "cool, but ain't never gonna happen" approaches, but they shouldn't be limited to what can be done today.

That's not the goal of this HIPE. The goal here is to have the introduction messages. This only shows the minimum required fields. The contents of *message* can be whatever is necessary to complete the setup. The problem is we haven't define exactly what this is. So this should be the "Dependent On" section that you suggest.
> 
> I'd prefer to see a complete HIPE for Establishing Connections, with perhaps a "Current Status" section that deals with what can be implemented in the short term. Ideally, there should also be a "Dependent On" section that points to other HIPEs/features needed.
> 
> It's not clear my suggestion is right (although it is how I understand the HIPE process). I think if we do go with the more pragmatic approach you are proposing, then we should back off on several of the other HIPEs that have been proposed - especially the Cross-Domain and DIDDoc Conventions ones, as they are contradictory to this HIPE - and this one has code.

Cross-Domain should work in conjunction with this HIPE. This HIPE does not cover how routing and discovery are made. Just the interactions needed to complete the setup for a new relationship in a secure forward secret manner.
> 
> Finally - I would add that this to me shows the urgency of getting DIDDoc support in at least indy-sdk and likely indy-node.

Agreed.

text/agent-connection-protocol/README.md
```diff
@@ -0,0 +1,171 @@
...
+The following message formats are used by the connection protocol.
+
+**Connection Offer Unencrypted**
+```json
+{
+    "@type": "did:sov:1234567890;spec/connection/1.0/offerunencrypted",
+    "content": {
+        "public_key": "<base58 encoded key>",
+        "message": {
+            "return_path": ["<url for each hop to return>"]
```
tplooker: Can you clarify or stipulate any further what is meant by this return_path featuring each hop to return?

mikelodder7: Sure. Since this is meant to be the bootstrap for wire messages, the routing mechanism can be anything at the beginning. Alice needs to indicate how the Bob can reach her. Bob indicated how to reach him by either having a public DID on the ledger or Bob sent Alice an out of band message like an email, SMS, QR code scan, whatever.

tplooker: @mikelodder7 thanks, can you provide an example of the form this might take, i.e the purpose and usage of each hop, by hop do you mean network or endpoint hop?

PR #47, opened 2018-10-16
FCP universal-codec HIPE

sklump:
Proposal for universal JSON-driven codec layer to build into indy-sdk.

Is this even a good idea? The process of design and implementation has raised some existential questions. Further discussion required to arrive at consensus regarding best way forward.

Signed-off-by: sklump <srklump@hotmail.com>


vimmerru:
In my vision SCHEMA should be the main source for the correct interpretation of parameters in credential, but not parameter value in credential. We don't have just strings and numbers, but usually we have names, lengths, colors, temperatures. 

Ideally application should understand temperature in Fahrenheit and Celsius systems as well as a length in meters and foots, and make business decisions based on this values.

As a result i suggest to have encoding information as a part of SCHEMA together with units and domain field bindings. Your proposal can be simpler to implement, but it opens the door for solutions that will be incompatible with future decisions.

sklump:
But these are semantics on the original values, orthogonal to the encoding regimen.

dhh1128:
This feels like an incomplete solution to me. It's trivial to say we'll encode complex numbers as strings in such-and-such a form--but this gains us almost nothing. We can't do zero-knowledge proofs on the values unless we encode them as 256-bit integers. So punting and saying "we'll just encode everything except integers as a string" is equivalent to saying "we'll just stop supporting zero-knowledge proving on anything except integers".

We need more flexibility and power than that.

Am I missing something?

sklump:
The current API takes a JSON-dump (string)
```
{
    "attr0": {
        "raw": "...", # must be a string or else indy-sdk raises exception
        "encoded": "..." # numeric decimal string
    },
    "attr1": {
        "raw": "...",  # must be a string or else indy-sdk raises exception
        "encoded": # numeric decimal string
    }
     ...
}
```
without specifying how to build the encoding, except that 32-bit integers must encoded to their natural  numeric (decimal) strings. What this HIPE discusses is to whether to standardize the encoding process, taking the responsibility into the indy-sdk, and how to build the encoding.

Since JSON encoding does not support complex numbers natively (among many other types), and they're not boolean, null, objects, or arrays, they would encode as strings as it stands today anyway, the complex number scenario is equally applicable to today's API.

There may already be a future enhancement to this end that I don't know about.

swcurran:
> This feels like an incomplete solution to me. It's trivial to say we'll encode complex numbers as strings in such-and-such a form--but this gains us almost nothing. We can't do zero-knowledge proofs on the values unless we encode them as 256-bit integers. So punting and saying "we'll just encode everything except integers as a string" is equivalent to saying "we'll just stop supporting zero-knowledge proving on anything except integers".
> 
> We need more flexibility and power than that.
> 
> Am I missing something?

@dhh1128 - added a comment to the related JIRA about this, but adding it here for completeness.  

The answer to Daniel's question on the HIPE is that we have nothing now so we have no chance at interoperability - every Dev is making up their own solution (as Stephen and Axel have both done). Using Stephen's solution would give Devs an appropriate API (encode with only raw data input, and decode) to use going forward, and we can add what's needed to make the encoding ZKP-friendlier in the future.  I think (not certain) Stephen put in a "version" mechanism in the encode so that we should be able to improve the capability later.  I'm guessing the more useful capabilities would require enhancements to the schema object - adding in claim metadata to make improved ZKP opportunities possible.

text/0021-universal-codec/README.md
```diff
@@ -0,0 +1,220 @@
...
+booleans, large integers, floating points, arrays, nested JSON objects)
+in favour of their 'raw' string representations.
+
+_Question:_ Is there a substitution attack available, in finding raw
+values, if decoding is ever required?
+
+_Question:_ Why even have encodings? Would there be any drawback in
+enumerating non-32-bit-integer attribute values per credential,
+`maxint32 + i` for non-32-bit-integer attribute value #`i` in
+any given credential? Once the encoding is in place in a signed
```
jovfer: I don't understand this question, especially in this section.
How Verifier will check the fact like some proof is created by signed Indy artifact but not by malicious Prover? Like malicious Prover can create valid proof by indy and replace raw value with something useful for him, but not signed by Issuer.

jovfer: Let me describe the principal points of current anoncreds implementation to be on the same page and remind some nuances for this discussion participants:

The math in IndyCrypto allows to issue some credential, signed by Issuer with integer attributes (Daniel has mentioned 256bits above, my feeling was 254bits, but it's not a question for the discussion)

IndySDK allows to store the math proof together with raw values just for usability.

On Verifier side IndySDK+IndyCrypto validate that math proof is correct. In other words it checks the fact: the long integer revealed by proof has been signed by Issuer.
For this moment, it's up to Verifier application to make sure that raw value (which will be used in UI most probably) is corresponds to integer value (confirmed by proof).

jovfer: > Why even have encodings? Would there be any drawback in
enumerating non-32-bit-integer attribute values per credential,
`maxint32 + i` for non-32-bit-integer attribute value #`i` in
any given credential?

So my answer here: we have to use some encoding known and approved by all sides (Issuer, Prover/Holder and Verifier) to perform proof issuance and validation (in theory Prover may not known the encoding for some edge cases).
The math will verify validity of integer representation, but mapping between raw and encoded value is subject to verification too.

TelegramSam:
@sklump @jovfer There are questions without responses here. Are these resolved elsewhere, or are there things that still need to be discussed?

sklump:
Encoding algorithms must be specified for interop somewhere, but this universal approach is not the way to do it.

jovfer:
@sklump please take a look on my comments above. I got the point that this pr is only about some universal approach and not about particular encoding, but still have unresolved questions

sklump:
> 
> 
> @sklump please take a look on my comments above. I got the point that this pr is only about some universal approach and not about particular encoding, but still have unresolved questions

The metadata in the encoding allowed for decoding back to original data types (e.g., 0.05, None, True instead of "0.05", "", "1"). This turns out not to be a requirement and so we can discard the whole tactic.

text/0021-universal-codec/README.md
jovfer: @sklump please take a look on this thread too. The commented question must not be in `Unresolved questions` section in my mind

jovfer:
@sklump please let me remind my question above https://github.com/hyperledger/indy-hipe/pull/47#discussion_r249528702

sklump:
Unfortunately I don't understand it either anymore. I think it's safe to ignore it.

jovfer:
But the "unresolved" question in the discussion above seems as question from person, who don't understand the CL ZKP completely. So it NOT safe to ignore it as it may confuse reader who try to understand the system.

sklump:
Let me rephrase the question.

What good is the encoding specification itself? As long as nobody can decode from encoded back to raw, what does it matter what encoding corresponds to what raw value? As long as the signature on the credential testifies that raw value #i corresponds to encoded value #i for operations, preventing substitution attacks, why do we care how the issuer encodes raw values? They may as well be ordinals (outside the int32 -> same int32 stricture).

There has to be a flaw in the above argument that comes from naïveté. If it is important, kindly correct it.

jovfer:
> As long as the signature on the credential testifies that raw value #i corresponds to encoded value #i for operations, preventing substitution attacks, why do we care how the issuer encodes raw values?

What do you mean by "signature" here? The math proof based on credential issued by some issuer keys?

sklump:
> 
> 
> > As long as the signature on the credential testifies that raw value #i corresponds to encoded value #i for operations, preventing substitution attacks, why do we care how the issuer encodes raw values?
> 
> What do you mean by "signature" here? The math proof based on credential issued by some issuer keys?

I think so: from the call anoncreds.issuer_create_credential(..., cred_values_json: str, ...), the cred_values_json maps raw attribute values to encoded values. The resulting credential should cement this relationship forever via some kind of signature, no? And once in place, does it matter how we got the raw:encoded value, as long as, within the credential, all distinct raw values map to distinct encodings?

jovfer:
> The resulting credential should cement this relationship forever via some kind of signature, no?

No. Sorry, I'm not able to answer in details now, please see https://github.com/hyperledger/indy-hipe/pull/47#discussion_r249531193 and chat me in Rocketchat if my explanation is not clear enough

sklump:
> 
> 
> > The resulting credential should cement this relationship forever via some kind of signature, no?
> 
> No. Sorry, I'm not able to answer in details now, please see [#47 (comment)](https://github.com/hyperledger/indy-hipe/pull/47#discussion_r249531193) and chat me in Rocketchat if my explanation is not clear enough

Oh! Thanks. I hadn't understood it properly.

At this point I believe everyone uses sha256(str(value)). In the incredibly unlucky event that a non-32-bit value hashes to a 32-bit int, von_anchor hashes again until it doesn't. I've proven exhaustively that no 32-bit int hashes to another 32-bit int (and this holds true for both big-endian and little-endian), so this process is guaranteed to terminate after at most one additional iteration.

Once a proper encoding guideline comes out, von_anchor will adopt it.

dhh1128:
@sklump This has languished for a long time. Do you still want to merge it? Now that we're using a more streamlined process, we could do so quickly...

sklump:
https://github.com/hyperledger/indy-hipe/pull/119 captures the essence of the thing and more. We can close this PR.

PR #48, opened 2018-10-22
Supporting multiple keys and endpoints for a DID

lovesh:


text/0021-did-doc/README.md
```diff
@@ -0,0 +1,108 @@
+- Name: did-doc
```
dhh1128:
```suggestion
- Name: enhanced-did-doc
```
(PR #37 already proposes some DID Doc conventions, so this extends rather than replaces it.)

text/0021-did-doc/README.md
```diff
@@ -0,0 +1,108 @@
...
+- Author: lovesh.harchandani@evernym.com
+- Start Date: 2019-10-18
+- PR: (leave this empty)
+- Jira Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Indy currently supports only 1 public key per DID. This HIPE proposes to add support for more than 1 public key per DID. Additionaly the keys can have one or more authorizations. The authorizations enable granular control over the DID doc like which keys can add new keys, which keys can remove which keys. This HIPE also proposes to add support for endpoints for a DID. Endpoints specify where the DID can be connected to. Endpoints can follow different protocols, HTTP, HTTPS, SMTP, even DID.
```
dhh1128: "Additionaly" --> "Additionally"

text/0021-did-doc/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+A DID can have several agents. It might want some of it's agents to be able to add new agent, some of them to remove existing agents, some with all possible authorizations. eg, an organisation (DID) can designate some of it's members (agents) to add new members, some to remove existing members, some to be able to do both.
```
dhh1128: "it's agents" --> "its agents"
"add new agent" --> "add new agents"
"some of it's members" --> "some of its members"

text/0021-did-doc/README.md
```diff
@@ -0,0 +1,108 @@
...
+A DID needs to let others know where they should send messages in case they need to communicate. eg, An organisation (DID) can declare that to communicate by email, send an email at foo@example.com, to reach by HTTP visit https://example.com, to send money by Bitcoin, send at a particular address.
+The addition of above 2 features brings Indy DIDs somewhat closer to W3C's DID spec.
+
+# Tutorial
+[tutorial]: #tutorial
+
+## Concepts
+1. Key reference: Any key added to a DID gets a reference unique in context of that DID. This reference is a monotonically increasing integer without any gaps, starting from 1. Hence the first key for the DID (during creation) is given a reference of 1. The next key gets a reference of 2. This reference is used to change the public key in case of compromise or otherwise or grant additional authorizations. Only the key with key reference 1 can be used to act as a role like Steward, TGB, Trustee, etc.
```
dhh1128: I'm not sure about "Only the key with key reference 1..." Isn't it the *DID* that's given the role of Steward? Maybe "DID_ROLE" should be a role given to zero or more keys for a DID?

text/0021-did-doc/README.md
```diff
@@ -0,0 +1,108 @@
...
+2. Authorizations: Any public key can possess any number of authorizations. These determine what actions the key can take. A key's authorizations can change over time. It is possible for a key to have no authorization at all.
+Following authorizations are possible:
+a. `ALL`: A special authorization; a key with this authorization can change the DDO however it wants. The key with reference 1 has this authorization.
```
dhh1128: I don't think this authorization is a good idea, because its meaning will change if we add new authorizations in the future. In order to assign what is "ALL" today, we should enumerate what "ALL" means.

text/0021-did-doc/README.md
```diff
@@ -0,0 +1,108 @@
...
+b. `ADD_KEY`: Add new keys to the DID. The newly added key can only have authorizations that the key adding it has. 
+c. `REM_KEY`: Remove any existing keys from the DID. Keys with this authorization can remove any key with any authorization, i.e. even a key with ALL authorization. This can be debatable hence mentioned in the last section of this HIPE.
+d. `MOD_KEY`: Update the key and/or it's authorizations. Any key can change its own public key irrespective of it having `MOD_KEY` authorization or not.
```
dhh1128: "it's authorizations" --> "its authorizations"

text/0021-did-doc/README.md
```diff
@@ -0,0 +1,108 @@
...
+e. `MOD_EP`: Add, remove or update endpoints.
+3. Endpoints: Endpoints are URIs where the DID can be interacted with. They can be HTTP, HTTPS, TCP, SMTP or even DID. The messages sent to a DID need to be encrypted. If the endpoint is not a DID then it needs to be specified which key to use with that endpoint. If it is a DID then the DID can be looked up to find a key with `tags` (see below for explanation) containing `MPROX`. There is an exception to this rule is the endpoint type `PAY`. This endpoint type is used to specify a payment address. This exception can be debatable hence mentioned in the last section of this HIPE.
+Each endpoint is given a reference similar to the way in which keys are given reference.
+4. Tags: A key can specify a list of strings called `tags` for any higher level application to parse. Currently only one tag is specified by Indy called `MPROX`. Only a key can update it's `tags`
```
dhh1128: "it's tags" --> "its tags"

text/0021-did-doc/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Authorization rules subtleties
+1. Any key can forfiet it's 1 or more authorizations. eg. if a key has `ADD_KEY` and `REM_KEY` authorizations, it can give up it's `REM_KEY` authorization even if it does not have `MOD_KEY` authorization.
```
dhh1128: "it's 1" --> "its 1"
"it's `REM_KEY`" --> "its `REM_KEY`"

text/0021-did-doc/README.md
```diff
@@ -0,0 +1,108 @@
...
+2. When a new key, say `subject` is added by another key, say `actor`, `subject` can only have authorizations which `actor` has. `actor` might decide to give less authorizations to `subject` but it cannot give more. eg. Key `k1` has authorizations `ADD_KEY` and `REM_KEY`, it adds a new key `k2`, now `k2` can at most be given `ADD_KEY` and `REM_KEY` by `k1`, `k1` might give just `ADD_KEY` or `REM_KEY` or no authorization at all but `k1` cannot give `MOD_KEY`.
```
dhh1128: The use of "actor" and "subject" here is confusing, since we said keys are referenced by a number like '1' or '2'. You should either call these keys '1' and '2', or you should be more explicit in your language, and not use notation that suggests words like "actor" and "subject" have a meaning in code: "When a new key, that we'll call [subject], is added..."

text/0021-did-doc/README.md
```diff
@@ -0,0 +1,108 @@
...
+3. When an existing key, say `subject`'s authorizations needs to be modified, they can only be modified by a key with `MOD_KEY` authorization. Moreover, this key, say `actor` cannot grant the key any more authorizations than it has. Though it can take away any authorizations. This can be debatable hence mentioned in the last section of this HIPE.
+4. A key with `MOD_EP` can add new endpoint or change any endpoint's value. Changing an endpoint's value to empty string removes it. 
+
+# Reference
+[reference]: #reference
+
+## Authorization representation
+Authorizations are represented as bitset in transactions. 
+- `ALL`: Bit 0
```
dhh1128: As I said before, i don't think ALL is safe, because its meaning changes.

text/0021-did-doc/README.md
```diff
@@ -0,0 +1,108 @@
...
+```
+{type: MOD_KEY, did: <subject did>, publicKeyRef: <public key reference>, authorizations: <bitset for authorization>, tags: <tags>}
+```
+5. A new transaction called `EP` needs to be added. This transaction is used to add, remove or update. 
+A transaction to add new endpoint looks like this
+```
+{type: EP, endpoint: <the URI>, key_ref: <Optional field. The key reference incase the endpoint is not a DID>}
+```
+On adding the endpoint, it is assigned an integer which is used to reference this endpoint. The integer is monotonically increasing in the context of the DID, starting from 1, conceptually similar to the monotonically increasing key reference.
+Incase when the endpoint is not a DID, reference to one of this DID's key is needed sending encrypting messages to this endpoint.
```
dhh1128: "Incase" --> "In case"
"one of this DID's key" --> "one of this DID's keys"

text/0021-did-doc/README.md
```diff
@@ -0,0 +1,108 @@
...
+{type: EP, endpoint: <the new URI>, ep_ref: <integer referencing the endpoint>, key_ref: <Optional field. The key reference incase the endpoint is not a DID>}  
+```
+
+The transaction payload to remove an endpoint looks similar to this.
+```
+{type: EP, endpoint: '', ep_ref: <integer referencing the endpoint>}
+```
+
+## Implementation
+Currently indy-node stores the DID information in 2 query layers, the state trie which is queried for client queries since they need a proof and the `IdrCache` which is queried for signature verification and role based authorization checks. To implement above features it is proposed to store all the DID data like keys, endpoints, etc as a JSON for the DID (hash of it) in the trie. The `IdrCache` should also store a JSON for the DID with key reference and key values both indexed. The endpoint references should be indexed as well. This will require a migration of the trie and `IdrCache` during deployment. Something to note is that the DID doc ouptuted by Indy-sdk has to be compliant with the DID doc in Sovrin DID method spec but the data organization on Indy nodes is independent of that.
```
dhh1128: Substantive comment: Whether we store the full JSON of a DID Doc or things that can be assembled into a DID Doc is an implementation detail too low-level for this HIPE, probably.

trivial comment: "ouptuted" --> "outputted"

text/0021-did-doc/README.md
```diff
@@ -0,0 +1,108 @@
...
+For the short term, keys and endpoints stored using `ATTRIB` transaction. A DID `abc` can have an attribute `abc:k*` to contain a list of all key references like `[1, 2, 3]` and then there are 3 more attributes, `abc:k1`, `abc:k2`, `abc:k3` with the value of each aattribute being the corresponding data like the actual key, authorization and tag. Similarly endpoints can represented using attribute.
```
dhh1128: I would like not to extend the ATTRIB txn if we can avoid it. Can we eliminate it entirely?

text/0021-did-doc/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+None.
+
+# Rationale and alternatives
+The HIPE looks divergent the DID specification by W3C but is neccessary for use cases for Indy.
```
dhh1128: Please enumerate the ways that we diverge, specifically.

text/0021-did-doc/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+# Prior art
+1. [The DID specification](https://w3c-ccg.github.io/did-spec/).
+2. [Sovrin DID method specification](https://github.com/sovrin-foundation/sovrin/blob/master/spec/did-method-spec-template.html)
+3. The transactions and authorization rules are similar to [Relationship State Machine HIPE (In review)](https://github.com/hyperledger/indy-hipe/pull/31)
```
dhh1128: Please reference the HIPE in PR #37.

text/0021-did-doc/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+1. Should keys with `REM_KEY` be allowed to remove keys with `ALL` authorization or should keys with `ALL` be removable by keys with `ALL` authorization? 
+1. Implementing more flexible authorization like 2 of the 5 keys have to agree to add a new key.
```
dhh1128: I want 1 or 2 sentences of comment about how to do this, in the text of the HIPE, so we know it can be done without doing violence to the current format. Basically, I'm concerned that when we introduce this idea, we'll have a breaking change in txn format or in JSON format, and I want to pre-empt that.

text/0021-did-doc/README.md
```diff
@@ -0,0 +1,108 @@
...
+1. Should keys with `MOD_KEY` be allowed to change authorizations of a key with `ALL` authorization? Do we need an authorization called `ADMIN` that is distinct from `ALL` and immune to such actions?
+1. Should `pay` be introduces as an endpoint?
```
dhh1128: "introduces" --> "introduced"

text/0021-did-doc/README.md
lovesh: `Maybe "DID_ROLE" should be a role given to zero or more keys for a DID?` Do you mean `Maybe "DID_ROLE" should be a authorization given to zero or more keys for a DID?`. The reason only 1 key reference was allowed was that the actions associated with roles should be performed rarely. This is true for current roles (and their current responsibilities) like Trustee, TGB,  Steward, etc. But i can imagine Stewards keeping different keys for node maintainence, different for voting for governance, for voting nodes to be validators for certain rounds, etc. So i am not entirely opposed to it either. Would like to hear from a few more people

text/0021-did-doc/README.md
lovesh: Would you like it if i renamed it to Admin? An entity or maybe more of them that can take arbitrary actions.

text/0021-did-doc/README.md
lovesh: I did not consider this to be low level as there are no higher layers of implementation than this. But i don't want to be too prescriptive either so can remove if others feel so

text/0021-did-doc/README.md
lovesh: I don't like it either. Just that it was proposed as a short term alternative but can get rid of it.

lovesh:
@dhh1128 I have resolved the spelling and grammatical errors. Apologies, there were too many

text/0021-did-doc/README.md
lovesh: Done

text/0021-did-doc/README.md
lovesh: Added a section called `More flexible authorization rules (for future)` for this

text/0021-did-doc/README.md
lovesh: Can we move to admin

text/0021-did-doc/README.md
lovesh: Updated.

text/0021-did-doc/README.md
dhh1128: I'm happy with what was added.

text/0021-did-doc/README.md
dhh1128: You referenced PR #31, not PR #37.

text/0021-did-doc/README.md
dhh1128: I'm not sure I like "admin" either. Do we want a privilege that means "any privileges we can imagine, no matter what they are"? Or do we want to simply enumerate the privileges we know about? I'm not sure. I want some others to comment.

text/0021-did-doc/README.md
dhh1128: No. The name has nothing to do with my concern. It's the semantics that are making me uncomfortable.

text/0021-did-doc/README.md
lovesh: Added

text/0021-did-doc/README.md
lovesh: I did, here https://github.com/hyperledger/indy-hipe/pull/37#issuecomment-432330523. My comment in PR 31 for something else

text/0021-did-doc/README.md
lovesh: The property i wanted was that key with `ALL/ADMIN` authorization should be able to make whatever change it wants in the DID doc.

dhh1128: Right. That's the very thing I'm uncomfortable with.

swcurran:
Sam Smith did a really interesting talk at the Fall 2018 IIW about a good way of handling key rotation. Notably, when creating a new key, generate two keypairs, one as the current key, a second as the "rotation" key. Then, when a rotation (update) is done, the "rotation" key becomes the current key (it's never been used before), and the new key is the new "rotation" key - not used until the next rotation.  The benefit is that if the current key is compromised without loss of the wallet, the attacker cannot rotation the key - they don't have the private key of the the "rotation" key.  Evidently this technique is used by others.  Check with Sam for a reference.

Might be worth considering if we are going to be updating the key handling anyway.

lovesh:
> Sam Smith did a really interesting talk at the Fall 2018 IIW about a good way of handling key rotation. Notably, when creating a new key, generate two keypairs, one as the current key, a second as the "rotation" key. Then, when a rotation (update) is done, the "rotation" key becomes the current key (it's never been used before), and the new key is the new "rotation" key - not used until the next rotation. The benefit is that if the current key is compromised without loss of the wallet, the attacker cannot rotation the key - they don't have the private key of the the "rotation" key. Evidently this technique is used by others. Check with Sam for a reference.
> 
> Might be worth considering if we are going to be updating the key handling anyway.

If i understand correctly, the DID doc will specify the "rotation key", eg. while creating a new DID doc, i generate 2 keys `k1` and `k2` and put both in the DID doc. Here `k1` will be my current key and `k2` will be my key after rotation. The DID doc registry (ledger, etc) will ensure that my next key is `k2` and nothing else. This is problematic as nobody can change my keys unless he holds my wallet. Currently it is proposed to give priviledges to others for key change which will be difficult with such a scheme.
Secondly key compromise without wallet seems less common since keys never leave the wallet so it was either a flaw in the cryptographic scheme (signature/ciphertext reveals key) or a side-channel.

text/0021-enhanced-did-doc/README.md
```diff
@@ -0,0 +1,160 @@
...
+A DID needs to let others know where they should send messages in case they need to communicate. eg, An organization (DID) can declare that to communicate by email, send an email at foo@example.com, to reach by HTTP visit https://example.com, to send money by Bitcoin, send at a particular address.
+The addition of above 2 features brings Indy DIDs somewhat closer to W3C's DID spec.
+
+# Tutorial
+[tutorial]: #tutorial
+
+## Concepts
+[concepts]: #concepts
+
+1. **Key reference**: Any key added to a DID gets a reference unique in context of that DID. This reference is a monotonically increasing integer without any gaps, starting from 1. Hence the first key for the DID (during creation) is given a reference of 1. The next key gets a reference of 2. This reference is used to change the public key in case of compromise or otherwise or grant additional authorizations. Only the key with key reference 1 can be used to act as a role like Steward, TGB, Trustee, etc.
```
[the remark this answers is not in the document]
lovesh: > require that agents be monotically increasing. In very complex domains (e.g., for a corporation), it may be desirable to authorize multiple keys in parallel. 

Keeping changes seen by the DID doc registry sequential makes it easier to reason about as  parallel changes can lead to ambiguity, eg. 2 parallel changes like revoke key `4` and an action by  key `4`; to resolve this we need to have an extra condition saying accept parallel changes by same author as he knows what he is doing but sequence changes by different authors. Note that the client can still submit a batch of changes but these changes are ordered in a list. The responsibility of ordering the actions is on the client and when the client does not care about the order like provisioing a bunch of keys, he creates the batch in any order.

> it may be desirable for a person to randomly assign key references with gaps in between them, to avoid the ability to fingerprint the agentspace of a person for correlation purposes.

I see we can let the client pick his say 8 byte id and the registry ensures these ids are unique per DID doc. But i understand the "fingerprint" argument. Do you mean that by looking at the monotonic id someone can see how many agents were created or before or after him? Or something else. This information can still be deduced since all actions to the DID registry are recorded with timestamp (in order). Or we think might not be true for all DID registries?

> Regarding the sentence, "Only the key with key reference 1..." -- 

You are correct, this is limiting. Authorization for each possible ROLE seems to be the right thing since your ROLE allows (authorizes) you to take actions. Also the option to create new role needs to be present. I think we should have some predefined auths like for adding/removing keys, endpoints and an authorization to create custom authorizations. The creator of an authorization sets up how that auth is granted, revoked, etc. This is similar to how EOS does it IIRC.

text/0021-enhanced-did-doc/README.md
```diff
@@ -0,0 +1,160 @@
...
+2. **Endpoints**: Endpoints are URIs where the DID can be interacted with. They can be HTTP, HTTPS, TCP, SMTP or even DID. The messages sent to a DID need to be encrypted. If the endpoint is not a DID then it needs to be specified which key to use with that endpoint. If it is a DID then the DID can be looked up to find a key with `tags` (see below for explanation) containing `MPROX`. There is an exception to this rule is the endpoint type `PAY`. This endpoint type is used to specify a payment address. This exception can be debatable hence mentioned in the last section of this HIPE.
+Each endpoint is given a reference similar to the way in which keys are given reference.
```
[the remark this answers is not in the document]
lovesh: Yes.

text/0021-enhanced-did-doc/README.md
```diff
@@ -0,0 +1,160 @@
...
+3. **Tags**: A key can be associated with list of strings called `tags` for any higher level application to parse. Currently only one tag is specified by Indy called `MPROX`. To add or remove tags, a key needs to have authorization. A key can forfeit any of its `tags`
+4. **Authorizations**: Any public key can possess any number of authorizations. These determine what actions the key can take. A key's authorizations can change over time. It is possible for a key to have no authorization at all.
+Following authorizations are possible:
+a. `ADMIN`: A special authorization; a key with this authorization can change the DDO however it wants. The key with reference 1 has this authorization.
```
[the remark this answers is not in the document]
lovesh: Ok, thanks. Will fix it?

text/0021-enhanced-did-doc/README.md
```diff
@@ -0,0 +1,160 @@
...
+2. **Endpoints**: Endpoints are URIs where the DID can be interacted with. They can be HTTP, HTTPS, TCP, SMTP or even DID. The messages sent to a DID need to be encrypted. If the endpoint is not a DID then it needs to be specified which key to use with that endpoint. If it is a DID then the DID can be looked up to find a key with `tags` (see below for explanation) containing `MPROX`. There is an exception to this rule is the endpoint type `PAY`. This endpoint type is used to specify a payment address. This exception can be debatable hence mentioned in the last section of this HIPE.
+Each endpoint is given a reference similar to the way in which keys are given reference.
+3. **Tags**: A key can be associated with list of strings called `tags` for any higher level application to parse. Currently only one tag is specified by Indy called `MPROX`. To add or remove tags, a key needs to have authorization. A key can forfeit any of its `tags`
```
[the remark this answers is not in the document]
lovesh: Sounds good. Will change

text/0021-enhanced-did-doc/README.md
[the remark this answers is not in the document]
lovesh: Would it be to complex if added another authorization called `REVOKE`?

vimmerru:
@lovesh @dhh1128 @ashcherbakov 

I have the following concerns:

1. I don't like idea of creation of custom transactions related to authorization/endpoints rules definition . All this authorization/endpoints staff is just part of DID Doc. Better alternative from my point of view is just define 2 transactions: 
    - To create DID Doc
    - To update DID Doc using delta

They can be just extended version of existing NYM transaction.

2. I don't like idea of definition any key identifiers/references that are against w3c spec. Each key in DID Doc already has identifier according to w3c spec. Why we don't use it. 

3. w3c spec suggests to use authorizations field to define authorizations rules. Why we don't use it? We can just add key role as additional field to authorization record and be completely w3c spec compatible. 

I suggest to look to this PR https://github.com/hyperledger/indy-sdk/pull/1487/files that describes alternative approach.

ashcherbakov:
I agree with Slava.
The only thing is that I would prefer not to modify existing NYM, but just add two new txns: DID_DOC (creation of new DID and corresponding did doc) and DID_DOC_DELTA (update of existing did doc).

text/0021-enhanced-did-doc/README.md
```diff
@@ -0,0 +1,160 @@
...
+- `ADMIN`: Bit 0
+- `ADD_KEY`: Bit 1
+- `REM_KEY`: Bit 2
+- `MOD_KEY`: Bit 3
+- `MOD_EP`: Bit 4
+- `ADD_TAG`: Bit 5
+- `REM_TAG`: Bit 6
+
+
+## New transactions
```
ashcherbakov: I would prefer not to modify existing NYM and a lot of custom txns,
but just add two new txns: DID_DOC (creation of new DID and corresponding did doc) and DID_DOC_DELTA (update of existing did doc). Ledger needs to be responsible for validation and parsing of DID_DOC_DELTA txns and merging it with the current DID_DOC (last state) to create the new state.

dhh1128:
Conversations on the topic of DID Doc support and microledgers are unfolding in several different places -- rocket.chat, slack, google docs, jira. Not all of the same people are in each place, and I didn't want to copy and paste a lot of text all over. So I put some thoughts in a doc. Please have a look, and please forward this on to other people if you think they should be aware:

https://docs.google.com/document/d/17qzu5jm2CIA32oArG4cRp7g0D_20cwUqB4sX4LIkodc/edit#

text/0021-enhanced-did-doc/README.md
```diff
@@ -0,0 +1,160 @@
+- Name: enhanced-did-doc
+- Author: lovesh.harchandani@evernym.com
+- Start Date: 2019-10-18
+- PR: (leave this empty)
+- Jira Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Indy currently supports only 1 public key per DID. This HIPE proposes to add support for more than 1 public key per DID. Additionally the keys can have one or more authorizations. The authorizations enable granular control over the DID doc like which keys can add new keys, which keys can remove which keys. This HIPE also proposes to add support for endpoints for a DID. Endpoints specify where the DID can be connected to. Endpoints can follow different protocols, HTTP, HTTPS, SMTP, even DID.
```
swcurran: I assume we're not limited to those protocols. - "Endpoints can follow different protocols such as HTTP, HTTPS, SMTP, even DID and others."

text/0021-enhanced-did-doc/README.md
```diff
@@ -0,0 +1,160 @@
...
+2. **Endpoints**: Endpoints are URIs where the DID can be interacted with. They can be HTTP, HTTPS, TCP, SMTP or even DID. The messages sent to a DID need to be encrypted. If the endpoint is not a DID then it needs to be specified which key to use with that endpoint. If it is a DID then the DID can be looked up to find a key with `tags` (see below for explanation) containing `MEDIATE`. There is an exception to this rule is the endpoint type `PAY`. This endpoint type is used to specify a payment address. This exception can be debatable hence mentioned in the last section of this HIPE.
+Each endpoint is given a reference similar to the way in which keys are given reference.
+3. **Tags**: A key can be associated with list of strings called `tags` for any higher level application to parse. Currently only one tag is specified by Indy called `MEDIATE`. To add or remove tags, a key needs to have authorization. A key can forfeit any of its `tags`
```
swcurran: There is no definition of the one magic key - `MEDIATE`. It sounds like it is perhaps mediator, which means this is not aligning with the DIDDoc Conventions HIPE that is happening in parallel.

text/0021-enhanced-did-doc/README.md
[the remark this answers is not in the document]
swcurran: I don't think the monotonically increasing key will work in conjunction with requiring support for multiple agents updating/transmitting the DIDDoc in parallel, possibly on a split network (an example I've heard).

text/0021-enhanced-did-doc/README.md
swcurran: This gets particularly difficult when we are trying to do this as a DID update that is interoperable across other DID networks (Veres 1, uPort, etc.). I think the simpler the better, so create and delta sounds much better to me as well.  I also wonder about just create and replace, but I get that people don't like the need to send the whole diddoc on each transaction.

text/0021-enhanced-did-doc/README.md
```diff
@@ -0,0 +1,160 @@
...
+The transaction for adding/updating/removing keys will not change apart from additional authorizations for policies.
+
+# Reference
+[reference]: #reference
+
+## Authorization representation
+
+Authorizations are represented as bitset in transactions. 
+
+- `ADMIN`: Bit 0
```
swcurran: Bits?  Really?  In 2019?  :-).  Seriously though - is this really necessary?

dhh1128:
This should have been merged long ago. Sorry. However, it has been superseded since it was raised. The basic concept is now reflected in work happening at DIF, W3C, etc. Closing.

PR #49, opened 2018-10-22
Wallet Storage Plug-in - Postgres Lessons Learned, Outstanding Issues

ianco:
This HIPE describes a common set of tools and framework for developing wallet storage plug-ins, including:

- Standard rust software components to facilitate building new plug-ins (in rust)
- Standard test suite for verifying correct implementation of the plug-in
- Standard interfaces and constants exposed from the Indy-sdk shared library

A Postgres plug-in has been developed to illustrate the proposed framework, and to be used to facilitate discussion of how the above points will be implemented.  The initial codebase for the Postgres plug-in is available in PR https://github.com/hyperledger/indy-sdk/pull/1238.


vimmerru:
Hi, In general i like the idea of providing better example for plugged wallet storage and corresponded test harness, but in details this HIPE and corresponded PR to indy-sdk repo (https://github.com/hyperledger/indy-sdk/pull/1175) breaks some our fundamental assumptions:

1. Libindy tests suite is intended to tests libindy (include pluggin interfaces), but isn't intended to test plugins or to be plugins test harness:
  - Plugin test harness must be targeted to test wallet storage functionality, but it may be just 2-3% of our test cases
 - Plugin test harness should guide plugin developer that ins't the case for very big libindy tests suite.
 - Libindy test suite was designed to provide reproducible results for some git commit and can rely on default storage and nullpay internals.
2. Libindy can't be responsible for loading plugins. How to link and load plugins is responsibility of an application. Each plugin just provides init function that calls corresponded register method in linked libindy. In a lot of supported platforms like iOS and Android libraries loading in runtime is impossible at all or have a lot of limitations.

I suggest you to consider my alternative proposal about this:

1. Wallet storage test harness should be dedicated native library with one method ```test_wallet_storage(storage_type, options)```
2. ```test_wallet_storage``` executes the short list of well documented test cases and generates report
3. Wallet storage test links libindy, test harness lib and wallet storage in the way he needs. Calls ```init``` function of storage and  ```test_wallet_storage``` after it.

ianco:
Thanks for your feedback!

For plug-in testing the reason I took the approach I did was that the unit tests were already available, and they covered exactly the level of testing provided to the default storage plug-in.  The only concerns I have with your suggested approach is that we'd wind up with a separate codebase that would need to be maintained, and may not be exactly in sync with what is being provided for the default wallet.

This is the same concern that is identified in tiket IS-881 (difference in behaviour between the default wallet and plug-in) - using the same set of unit tests for both the default and plug-in behaviour would alleviate that problem to some extent.  Note that when I ran the libindy wallet tests against the postgres plug-in it identified errors in the libindy plug-in module (see updates in https://github.com/hyperledger/indy-sdk/pull/1247/files#diff-3ef068287d272575d182d10ff4517073) as well as a gap in the functionality provided to marshall/un-marshall the search query between libindy and the plug-in (see new method in https://github.com/ianco/indy-sdk/blob/postgres_plugin/samples/storage/storage-postgres/src/postgres_storage/language.rs#L172).  Both of these would have been a challenge to solve without leveraging the libindy unit tests.

Maybe an approach is to do as you suggest (separate wallet test harness), but have this harness incorporate the libindy test suite somehow.

Overall I think the plug-in test suite needs to provide a "robust" set of functionality, consistent with the level of tests that we have available for the default wallet, and it should be maintained, so that any improvements to the default wallet testing (or bug fixes etc.) are also reflected in the plug-in test harness.  I agree that one purpose of the test harness is to guide developers, but another is to make sure that the plug-in acutally works.  Since all plug-ins should support the same set of functionality (and it should be the same as the default wallet), using one common test suite makes sense to me.

ianco:
Question regarding the rust wrapper - is this what we should be using for external rust applications (such as the CLI or storage plug-ins) to access libindy?  Are there some examples or documentation available?  I thought I saw somewhere that it was considered "experimental" so not sure if we should be incorporating this yet or not ...

ianco:
There is a lot covered in this HIPE and the related PR 1247, so I've created JIRA's for each topic/issue so that they can be addressed individually.  I can create separate PR's if necessary.

The higher priorities are the issues with the SDK (IS-1045, 1048) and duplicated/shared code (1046, 1047) that prevent us from deploying into production (we are currently running against our own indy-sdk fork).

Lower priority issues are those that will provide value to the community but don't block us right now.

Jira Issue(s):
    - IS-1045 - plugged wallet return error codes (related - IS-804?)
    - IS-1046 - marshall/un-marshall for query arguements for wallet storage search
    - IS-1047 - shared code for plugged wallets (rust)
    - IS-1048 - add register_wallet_storage to python wrapper
    - IS-1049 - update CLI to be able to register/load new plugged wallet types (related IS-1041?, IS-947 for VCX)
    - IS-1050 - re-factor in-mem plugged storage example to shared library
    - IS-1051 - test harness for plugged wallets (e.g. unit test "shim", stand-alone test harness, CLI integration) (related - IS-881)
    - IS-1052 - document how to build plugged wallet (libindy api, rust wrapper, shared code, etc.) (related IS-1037, IS-1032, IS-1031, IS-1030?)
    - IS-1053 - implement plugged storage for Postgres database

ianco:
Re IS-1047 I've added a suggestion to the JIRA ticket, I'll update the README in the HIPE.

ianco:
Updated to include reference to common (shared) code between libindy and wallet storage plug-in(s).

vimmerru:
> Updated to include reference to common (shared) code between libindy and wallet storage plug-in(s).
> In addition there are dependencies on some Indy-sdk code, these have been re-factored into a separate shared common library:

I spend some time for thinking and now i am not sure that introduction of libindy-common this way will help us a lot:

1. Interfaces inside of libindy-common are non-public. They can be very dynamic and often part of refactoring. For example, we plan to switch to more modern error handling approach soon.
2. Moving them to local shared crate will make them public and cause that any libindy refactoring will affect all dependent projects. I don't want to break samples during any refactoring of libindy.
3. Only storage plugins samples are main consumers of proposed shared crate now will consume only smart part of shared interfaces.
4. It can make vision that you have to use Rust to write libindy plugins.
5. It will be hard to consume libindy-common outside of libindy repo for creation of 3d party plugins. It will cause copy\paste anyway.
6. It will significantly change libindy development ergonomics. You will need to open separate project in IDE and maintain proper versioning.

As a result i suggest to copy paste minimum functionality you need and remove libindy-common at all.

vimmerru:
@ianco 

We can consider creation of separate repo called something like libindy-wallet-storage-boilerplate and make them as community maintained project if you still really want to avoid any copy-paste.

vimmerru:
@ianco 

I have feeling that having both "in-memory" and "postgress" plugins as a samples is a bit too much as we need to support more code in core project. Also "in-memory" plugin was designed as a test tool only and it isn't a good example from my point of view. It doesn't implement full wql and breaks some core assumptions about wallet persistence. I suggest the following:

1. Keep "in-memory" inside of libindy/tests as we have it now. 
2. Use only "postgress" plugin as a sample. 

Ideally i want to have "postgress" plugin in a dedicated repo maintained and published by community member outside of the core team. Libindy can have the link to this repo in the root Readme.md

vimmerru:
@ianco 

> As described above, a new command has been added to the CLI to register a new plug-in storage. Once registered, this can be used to create and open a new wallet, and supports all wallet functions.

CLI already supports loading any king of wallet storage plugins. You don't need to provide anything new. How it is designed to work:

1. Each libindy plugin should provide only one public function without params. Usually it is called *{plugin_name}_init()*. This function internally calls register_wallet_storage_type or register_payment_method. Corresponded callbacks usually mangled and doesn't available for direct call.
2. Application calls *{plugin_name}_init()* at the beginning and starts usage of registered storage types or payment methods.
3. cli provides command

```
indy> load-plugin library=<library-name> initializer=<init-func-name>
```

Corresponded design is here: https://github.com/hyperledger/indy-sdk/tree/master/doc/design/006-cli-plugins

For example of implementation you can look to libnullpay

vimmerru:
@ianco 

> Sharing database connections in a multi-threaded environment - Postgres connections cannot be shared between threads (in rust), so in the Postgres plug-in they are wrapped in a connection pool. This has not been fully tested, and there are potential stability issues (testing is on-going)

libindy uses one execution thread to handle all storage operation and in the future it will be at least one thread per storage. As a result you can remove the most of mutexes and implement lazy fetching of records without big problems.

PR #50, opened 2018-10-23
FCP HIPE: support multiple key types

lovesh:


text/0021-keys/README.md
```diff
@@ -0,0 +1,82 @@
...
+2. Keys with size greater than or less than 32 bytes. For these keys, read the bytes before the colon, i.e `:`; this is the prefix which determines what kind of key follows after `:`. The prefix is at most 7 bytes and the key is at most 64 bytes. (The reason is described under Reference).  
+
+The keys are always encoded as base58.
+
+```
+<32 byte ed25519 or Curve 25519 key, base58 encoded>
+```
+------------- OR --------------------------------
+```
+<at most 7 byte prefix>:<at most 64 byte key, base58 encoded>
```
dhh1128: How would we support an RSA key that is 4096 bits? Are you assuming the only type of crypto we will deal with is elliptic curve?

text/0021-keys/README.md
```diff
@@ -0,0 +1,82 @@
...
+- Author: lovesh.harchandani@evernym.com
+- Start Date: 2018-10-22
+- PR: (leave this empty)
+- Jira Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Supporting more than 1 key type (secp256k1, etc) for signing and encryption in Indy. This implies on-ledger and off-ledger usages. Different keys are represented by prefixing them with a fixed size identifier. 
+This HIPE does not address keys used for BLS signatures.
```
dhh1128: Why is it okay to ignore BLS signatures? If we implemented multi-sig among agents of a given user (e.g., 2 of 5 must sign to be able to add a new key on a microledger), would we need BLS support?

text/0021-keys/README.md
```diff
@@ -0,0 +1,82 @@
...
+[summary]: #summary
+
+Supporting more than 1 key type (secp256k1, etc) for signing and encryption in Indy. This implies on-ledger and off-ledger usages. Different keys are represented by prefixing them with a fixed size identifier. 
+This HIPE does not address keys used for BLS signatures. 
+This HIPE is not proposing to support any specific curve.
+
+# Motivation
+[motivation]: #motivation
+
+Indy currently uses EdDSA for signing using ed25519 keys. For encryption, it uses X25519 for key exchange (ECDH) using Curve25519 to derive a shared secret and then XSalsa20+Poly1305 for authenticated symmetric encryption. That leaves out a large chunk of users, developers and ecosystems which rely on the secp256k1 curve; the curve used by Bitcoin, Ethereum and their many forks. Moreover the existing ed25519 and Curve25519 curves might be vulnerable in the future. Hence we want Indy to be flexible to support future cryptography.
```
dhh1128: "; the curve used" --> use parens: " (the curve used ...)" (a semicolon sets off an independent clause)

text/0021-keys/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+# Tutorial
+[tutorial]: #tutorial
+
+There are 2 kinds of keys in Indy (excluding BLS keys)
+
+1. Keys with size 32 bytes. These are either ed25519 keys if used in a signing context or Curve 25519 keys if used in encryption context.
```
dhh1128: Are there any other elliptic curves that need exactly 32 bytes? What about Goldilocks?

text/0021-keys/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+They might do further checks to validate the provided key.
+
+# Reference
+[reference]: #reference
+
+This is a non-breaking change as the exisiting keys will continue to be supported. No migration of any sort is needed.
+
+To avoid someone sending large keys and putting junk on the ledger, the prefix is maximum of 7 bytes and the key is at most 64 bytes. Unless we have an accurate fees mechanism that accounts for byte size of transaction and a "storage rent" mechanism, such a guard is needed.
+The reason for chooising 7 bytes for prefix is that it allows `ed25519` and `curve25519`. Larger words like `secp256k1` can be shortened to `k256` which is their popular short name. The 64 byte limit for keys comes from a few facts;
```
dhh1128: typo, "choising" --> "choosing"

text/0021-keys/README.md
```diff
@@ -0,0 +1,82 @@
...
+a. Currently supported ed25519, Curve25519 and anticipated curve secp256k1 have 32 byte keys.
+b. Popular NIST curves NIST P-256 and NIST P-384 fit under 64 bytes.
+c. Safe curves like Ed448 and Curve511187 are also under 64 bytes.
```
dhh1128: Ah, here you're talking about Goldilocks. Ignore my comment above.

text/0021-keys/README.md
lovesh: Yes, i should mention this in HIPE

text/0021-keys/README.md
lovesh: Yes

text/0021-keys/README.md
lovesh: Mentioned

lovesh:
@dhh1128 Renamed to `key-types`

vimmerru:
I am ok too. Indy SDK already supports key prefixes in described way.

lovesh:
@dhh1128 @vimmerru Thanks. I will wait for review from @ashcherbakov

text/0021-key-types/README.md
```diff
@@ -0,0 +1,84 @@
...
+1. A prefix greater than 7 bytes is found.
+2. A key with size (excluding prefix) greater than 64 bytes is found.
+3. A key with a unsupported prefix are found.
+
+They might do further checks to validate the provided key.
+
+# Reference
+[reference]: #reference
+
+This is a non-breaking change as the exisiting keys will continue to be supported. No migration of any sort is needed.
```
ashcherbakov: The key type is already an explicit part of request signature which is stored together with a transaction on the ledger.
See https://github.com/hyperledger/indy-node/blob/master/docs/transactions.md#common-structure
Do you expect that we use prefix (as part of `value`) instead of `type` field? It may mean that we will need migration of data on the ledger.

lovesh: The current format assumes that all signatures in a single txn are same as the `type` field is outside, we can get rid of the `type` field. Regarding the migration, it should only be removal of `type` field from existing txns, right?

ashcherbakov: Yes, but even such a small change means migration of ALL txns on the Ledger that can be undesirable.

text/0021-keys/README.md
mikelodder7: Speed is not always top priority. I know in some instances that size is more important and BLS sigs are 1/2 the size of EdDSA. This comment is mostly for a note but keep this in mind going forward. Some micro controllers for IoT might want BLS sigs. Lets not restrict ourselves if we can avoid it.

text/0021-keys/README.md
mikelodder7: Goldilocks is bigger at 56 bytes.

kdenhartog:
looks like @lovesh might have to follow up with @ashcherbakov before we move this to FCP.

text/0021-key-types/README.md
dhh1128: @lovesh and @ashcherbakov Have you come to consensus on this open thread? If so, this PR can quickly be merged.

ashcherbakov: I agree that the proposed format is more flexible. 
If we OK that we will have to support the old format forever, then we can avoid migration.

lovesh: I vote for migration

ashcherbakov: Unfortunately, this is something we may not always control. SF may be against migration.
But in general I'm in favor of having the type for every signature value.

vimmerru:
@lovesh @dhh1128 @ashcherbakov 

I have the following concerns:

1. I don't like idea of creation of custom transactions related to authorization/endpoints rules definition . All this authorization/endpoints staff is just part of DID Doc. Better alternative from my point of view is just define 2 transactions: 
    - To create DID Doc
    - To update DID Doc using delta

They can be just extended version of existing NYM transaction.

2. I don't like idea of definition any key identifiers/references that are against w3c spec. Each key in DID Doc already has identifier according to w3c spec. Why we don't use it. 

3. w3c spec suggests to use authorizations field to define authorizations rules. Why we don't use it? We can just add key role as additional field to authorization record and be completely w3c spec compatible. 

I suggest to look to this PR https://github.com/hyperledger/indy-sdk/pull/1487/files that describes alternative approach.

ashcherbakov:
I think Slava's comment is more related to https://github.com/hyperledger/indy-hipe/blob/d435e76795f0ac5f990d623a16a73787bd83f3f1/text/0021-enhanced-did-doc/README.md, but I agree with Slava.
The only thing is that I would prefer not to modify existing NYM, but just add two new txns: DID_DOC (creation of new DID and corresponding did doc) and DID_DOC_DELTA (update of existing did doc).

dhh1128:
Conversations on the topic of DID Doc support and microledgers are unfolding in several different places -- rocket.chat, slack, google docs, jira. Not all of the same people are in each place, and I didn't want to copy and paste a lot of text all over. So I put some thoughts in a doc. Please have a look, and please forward this on to other people if you think they should be aware:

https://docs.google.com/document/d/17qzu5jm2CIA32oArG4cRp7g0D_20cwUqB4sX4LIkodc/edit#

dhh1128:
The google doc that I linked in my previous comment is fine, but we've now matured the conversation to the point where it should be considered "resolved".

dhh1128:
This should have been merged long ago. Sorry. However, it has been superseded since it was raised. Closing.

PR #53, opened 2018-10-31
Update Message Type HIPE to include guidelines

TelegramSam:


kdenhartog:
Propose to merge this. It's stale after 26 days. If we want to stay strict to the FCP procedure, I suggest we keep this short.

dhh1128:
We don't have to go through a new FCP period for HIPES that have already been merged and that are just getting clarifying tweaks. I'm going to go ahead and merge it.

PR #56, opened 2018-11-08
Proposed HIPE: Change the HIPE process to not be PR-centric

swcurran:
This is a proposed HIPE to change from the current PR-based HIPE incubation process to one based on merged PRs in different folders based on the status of the HIPE. Challenges with the existing process are presented and a HIPE life-cycle process is proposed.


dbluhm:
I agree that GitHub's concept of PR's is not an ideal solution to the HIPE process but I think the fact that this is how many other open source communities operates is significant and perhaps understated. This is the sort of structure I would expect to see when I am getting to know a new (large) open source project.

I think it's also beneficial to consider who the consumers and creators of HIPEs are. If the primary creators are not technically trained, the PR process is definitely a burden and other systems may be preferable. Same with consumers. But in general I think these HIPEs are intended to be read and consumed by maintainers and the technical contributors of the community. That being said, another system could still be conducive to those groups.

text/new-hipe-process/new-hipe-process.md
```diff
@@ -0,0 +1,121 @@
...
+* final-review-hipes
+* accepted-hipes
+* rejected-hipes
+* archived-hipes
+
+The README.md will be updated to reflect the new hipe process.  Existing HIPEs will be merged and immediately moved into the appropriate folders.
+
+## Propose a new HIPE - Creator
+
+Fork and clone the `hyperledger/indy-agent`.
```
dhh1128: wouldn't this be `hyperledger/indy-hipe`?

text/new-hipe-process/new-hipe-process.md
```diff
@@ -0,0 +1,121 @@
...
+
+Submit a PR for the new HIPE.
+
+Wait nervously for feedback and as received, update the HIPE.
+
+## Receive a new HIPE - Admin
+
+The repo admin reviews briefly the HIPE to make sure that the HIPE follows the rules for a new HIPE - that it is properly formatted, a JIRA exists, perhaps that the submitter is known (and connect with the HIPE creator if not), the DCO is included, etc.  Assuming so, the Proposed HIPE is merged.
+
+Ideally, the Admin or the Creator would announce the new HIPE on rocketchat.
```
dhh1128: I'd like to specify who does this (admin vs. creator; having either/or means nobody will do it), and specify which channel on rocketchat. Also, I suggest it be cross-announced on the mailing list.

text/new-hipe-process/new-hipe-process.md
```diff
@@ -0,0 +1,121 @@
...
+
+# Summary
+[summary]: #summary
+
+This HIPE defines a new HIPE process that reduces the use of Pull Requests (PRs) to manage the life cycle of HIPEs.  Rather than long-lived PRs, folders are used to manage HIPEs, making them much more accessible to creators and reviewers.
+
+# Motivation
+[motivation]: #motivation
+
+The current HIPE process is, quite frankly, a pain in the butt for most people. The steps are convoluted, or it requires far more GitHub knowledge than should be needed. The primary challenges to the current approach include:
```
dhh1128: Although I sympathize, let's reword "pain in the butt" to something like "frustrating experience" so it sounds more professional. :-)

text/new-hipe-process/new-hipe-process.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Provide Feedback - the Community
+
+The community should review the HIPE and provide feedback through the JIRA or by doing a PR against the HIPE - possibly with changes, possibly just with embedded comments. Another option for line-by-line feedback would be to comment against the commit (or perhaps there is a better in-GitHub way?).
+
+An example in-HIPE comment might be:
+
+> From: Stephen Curran - Hey, this idea is not the best way to provide feedback. What about using GitHub Issues for commenting?
+
+If PRs are submitted against the HIPE, the Admin would simply ensure that the comments are reasonable in following the process - not if they are good or bad. Admins would also break PR loops where several people are repeatedly adding/removing each other's changes.
```
dhh1128: This doesn't sound practical to me. All PRs against a proposed HIPE are automatically merged? How can someone who shows up later know what's the core idea and what's a counterproposal to it?

text/new-hipe-process/new-hipe-process.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Accept the HIPE
+
+A HIPE in the `final-review-hipes` folder uses the same "Promote" process above. The only difference is that prior to submitting the PR for the HIPE, the Creator should add a number to the HIPE folder (e.g. `0099-new-hipe-process`) by incrementing the number.  Any collisions on numbers will be resolved by the Admin merging the PRs.
+
+The Creator should announce the acceptance of the HIPE on the appropriate rocketchat channel.
+
+## Archive the HIPE
+
+A HIPE that is obsolete can be moved by a contributor by moving the HIPE folder to the `archived-hipes` folder. This would usually occur as a byproduct of the acceptance of a new HIPE or in conjunction with a request from the Indy-Maintainers group. The Admin reviewing the HIPE should make sure that it is appropriate to move the HIPE before merging the request.
```
dhh1128: Once HIPEs have been approved, I think we need to treat their URIs as permalinks. We can still use the archiving idea, but we need to leave a redirect page in the old location so people who come looking end up finding the content but knowing about the archive status.

dhh1128:
Generally, I love this. Let's make it happen. See my notes about a couple subtleties.

swcurran:
> I think it's also beneficial to consider who the consumers and creators of HIPEs are. If the primary creators are not technically trained, the PR process is definitely a burden and other systems may be preferable. Same with consumers. But in general I think these HIPEs are intended to be read and consumed by maintainers and the technical contributors of the community. That being said, another system could still be conducive to those groups.

I think this is why we want to do this. If those with a quasi-technical bent that base some of the discussion on, for example, business value vs. technical purity are excluded from the conversation, the viability of the project is reduced.  I've seen this tons of times in Open Source projects that think the only audience is the technical contributors.

text/new-hipe-process/new-hipe-process.md
swcurran: How about this.  The archive process would move the full folder to the "archived-hipes" folder, and a new folder+document would remain in the permalink location saying "This HIPE has been archived. The original HIPE document can be found here <Link to Archive>".

Does that work?

text/new-hipe-process/new-hipe-process.md
swcurran: I don't think it matters.  The "core idea" is what the community thinks is the current state of the proposed HIPE.  That says, worst case they can look a the history.

For example, in the case when you sent a pull request to my fork and I merged it, no one could tell (except by history) what you wrote and what I wrote.

text/new-hipe-process/new-hipe-process.md
swcurran: Corrected.

text/new-hipe-process/new-hipe-process.md
swcurran: Changed all instances to have the Creator do the notifications and to post to the Mailing List.

text/new-hipe-process/new-hipe-process.md
swcurran: Updated to reflect the changed approach that retains the permalink.

text/new-hipe-process/new-hipe-process.md
swcurran: Reworded

kdenhartog:
I'm in favor of giving this a shot if it accomplishes these two goals.

1. it's easier and faster for **current** maintainers and contributors to accurately discuss improvements to the project.

2. This helps to get _high quality_ code merged and In production faster. 

I believe this HIPE process can assist with #1 and nudge us towards #2, so I'm in favor.

With regards to the concern about commenting, I think it's worth trying to use GitHub issues to asynchronously track ongoing discussion about proposed HIPEs. We can use tagging to specific the HIPE being discussed and it's status.

I'd like to point out the somewhat obvious, but sometimes overlooked point, that as our community grows we may find this process lacks certain qualities at scale. If that occurs then we can address those issues in the future rather than wasting time bikeshedding on them now.

TelegramSam:
Is there a benefit to changing the document name away from README.md in the hipe folder? I like the consistency/dry nature of the way it is, but open to that change if it brings benefit.

swcurran:
>           Is there a benefit to changing the document name away from README.md in the hipe folder? I like the consistency/dry nature of the way it is, but open to that change if it brings benefit.

Good point - I've changed the reference in the text.  I just realized I also need to change the name of the file I've been editing and will do that.

nage:
I'm hoping we can keep our process in sync with what is generally used for the Rust RFC process here in Indy as well as in Sawtooth and now Hyperledger Ursa.  As such, I posted this PR to #sawtooth-core-dev and asked for feedback and got this, "my initial thought is that the feedback mechanism would be worse with the proposed change and since the entire point is to gather feedback and respond to it, probably not worth the trade-off".  I am opposed to a change that would make our process alien to folks familiar with how Rust or Sawtooth operate, though I think we can support some deviation.  I like the changes proposed to add status folders for tracking the files in the repo and allowing PRs to propose renaming files to the new status, but would like to preserve some relationship with the PR and review process.

jljordan42:
I think the question here isn't the mechanical merits of one process or another as that basically boils down to preference.

An important question I think that needs to be answered is who is the core audience of the HIPEs? 

A perhaps relatively small group of developer types or a potentially large audience of technical/business people that are trying to understand and may adopt Indy technology, others?

I think that considering that adoption of Indy technology by a global community is a goal .. there should be some way for non-developers to do some digging into HIPEs .. I know in my time in telecom that I would refer to RFCs from time to time.

My opinion is that developers will be more able to adopt to modest differences in these processes and perhaps this approach will be adopted by the other projects even .. who is to say that we can't have ideas that are adopted by other.

Alternatively .. if there was a way with a script or something that could tell more business oriented folks what HIPEs are the approved, pending, etc, and have links that take them to the comments and the current latest version then that could work as well ... I don't know GitHub enough if such a thing is possible .. something like a HIPE status page ... 

HIPE (with link to latest), Status, Comments ... with links etc

swcurran:
I don't think that adding a roadmap document would be that helpful.  Rather than decreasing the admin overhead of managing the HIPEs, this would increase the admin role to maintain the README as PRs are received and merges are done.  Or the README would be communally managed, which increases the complexity of branch management (edit the HIPE in a branch, edit the global README in master). I don't think the numbering change is particularly significant - nice to have, but does not meet the goal of making collaboration easier.

We've had a couple of people weigh in with split support, but I'm not sure how we get this moved to a terminal state.  Unlike other HIPEs, this affects everyone, so it should not be moved forward without the view of the team.

I don't find compelling the argument of doing it like others have done in the past, nor to be consistent with others who aren't part of the project want.  To me it comes down to the simple question of doing we think we will get more and better input with the existing approach or the proposed one.

dbluhm:
I'm not convinced that the proposed changes would provide an appropriate means of commenting on HIPEs. What is proposed here would appear to result in comments being spread out among issues, comments on commits, and in PRs against the repo (please correct me if I'm wrong). I agree still that the PR process isn't ideal but I'm not sure the alternative suggested here would really improve on the current system. I do really like the idea of a Jekyll site for the repo; I think that provides some interesting opportunities no matter what system we end up with.

dbluhm:
There is a new feature in GitHub (still in beta?) that allows people commenting on PRs to suggest changes that the PR Author can then merge as a new commit to the PR. This, I think, addresses some of the pains you've mentioned here.

swcurran:
The beta feature that Daniel mentioned: https://help.github.com/articles/incorporating-feedback-in-your-pull-request/  - give it a try in this PR so I can try to incorporate the changes.

That would certainly be a help.

text/new-hipe-process/README.md
```diff
@@ -0,0 +1,123 @@
+- Name: new-hipe-process
+- Author: Stephen Curran (swcurran@cloudcompass.ca)
+- Start Date: 2018-11-08
+- PR: 
+- Jira Issue: 
+
+# Summary
+[summary]: #summary
+
+This HIPE defines a new HIPE process that reduces the use of Pull Requests (PRs) to manage the life cycle of HIPEs.  Rather than long-lived PRs, folders are used to manage HIPEs, making them much more accessible to creators and reviewers.
```
swcurran:
```suggestion
This HIPE defines a new HIPE process that reduces the dependency on Pull Requests (PRs) to manage the life cycle of HIPEs.  Rather than long-lived PRs, folders are used to manage HIPEs, making them much more accessible to creators and reviewers.
```

dbluhm:
Looks like for the suggestion feature the DCO sign-off is not automatically added. This could make it a little less convenient.

swcurran:
:-(

PR #60, opened 2018-11-13
propose HIPE: message tracing

dhh1128:


khagesh:
General comment: IMO we should be utilizing open source standards to do tracing in systems. One good example for such standard is opentracing. Lots of people will immediately understand what to do and implement it easily. We already have good tools on top of standards such as zipkin, jaeger, etc.

text/message-tracing/README.md
```diff
@@ -0,0 +1,240 @@
...
+  `"(forwarded to did:sov:1234abcd#4)"`.
+
+##### 
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Tracing makes network communication quite noisy. It imposes a burden on message handlers.
+It may also incur performance penalties.
```
khagesh: It also opens system for Application level DOS attacks. For example: I can send a message via an agency and purposefully send trace url which is very slow to respond or network on trace receiver is throttled to receive data slowely. Combining this along with deep nested messages and have trace on every layer will essentially, very easily in just few hundreds, fill up whole connections quota and bandwidth of servers running agency.

text/message-tracing/README.md
```diff
@@ -0,0 +1,240 @@
...
+* `@type`: Should always be `"did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/tracing/1.0/trace_report"`,
+   or some evolved version thereof. Required for version control and to support trace sinks
+   that process other HTTP payloads as well.
+* `for-id`: The ID of the message that the handler is looking at when it composes the
+   trace report. Required.
+* `handler`: A string that identifies the handler in a way that's useful for troubleshooting purposes.
+   For example, it might identify a particular agent by DID+keyref, or it might be a friendly
+   string like "iPhone" or "AgentsRUs Cloud Agent, geocaching extension v1.3.7". Optional but
+   encouraged.
+* `elapsed-millis`: How many milliseconds did the handler have this message before composing
```
khagesh: It would be better if we could use `span` concept from open tracing standards.

dhh1128:
@khagesh -- Regarding the use of opentracing, I would say that this is an interesting idea. However, I am not sure it applies completely. Opentracing spans are about capturing units of *execution* -- that's why there are spans that overlap, and names of operations that need to be done. The goal in opentracing's world view is to follow how logic gets executed in a distributed application.

Here, we are talking about something that may cross application boundaries. When a credential is issued, some of the work that happens is done by a distributed application on the issuer's side, and some of it is done by one or more distributed applications on the holder's side. What we are trying to trace is not so much *execution within an app* -- it is more *how messages get sent and received* across a context bigger than one application controlled by one user or one stack.

We could perhaps use opentracing despite this mismatch. I would welcome suggestions about how to apply its principles to this slightly different problem.

text/message-tracing/README.md
dhh1128: Agreed.

text/message-tracing/README.md
dhh1128: Please see my larger comment about opentracing.

khagesh:
@dhh1128 I think opentracing also works in similar way that we are proposing here. They have a server which is listening for messages and that endpoint is used by all services, apps or clients and send trace/span data to server url. Span can be used for both performance monitoring and tracing a message from one service to another or to another edge to edge, these clients needs to report using same message ID.
If we see zipkin, then they provide a nice UI as well to visualize this incoming data and we can filter out on message ID or application/service

text/message-tracing/README.md
TelegramSam: I think some wording about the application of tracing would help. Specifically, this might be most often used in a dev environment or within the production environment of a single company. Other parties can choose to participate in the trace process and may require permission, payment, or impose strict throttling limits.

TelegramSam:
I am also feeling a mismatch between the goals of opentracing and what is being proposed here. Opentracing feels very much focused on the operations within a system, including what language is being used. This is mostly concerned with the routing of messages between systems. Though the basic form is similar (submit trace report to a server), the goals appear very different.

I would also love more information about any misconceptions I have. Perhaps a screenshare/quick demo would help.

dhh1128:
@khagesh The difference in thinking shows up in the assumptions about how the mechanism is configured to begin with. In opentracing, you have to instrument your code. They distribute libraries in 9 languages to help with this intrumentation. I believe the developer goes into all the functions that they want to be traceable, and add calls to the tracing function.

What I am proposing here requires no code instrumentation and no prior agreement or configuration. We simply agree that *message* (not *code path*) tracing can be requested in a particular way, and if parties support it, they honor the trace requests. They don't do this by scattering trace statements all through their code; they simply go into the one place in code where messages are received and sent, and implement tracing once, there.

Perhaps that one place of implementation could use opentracing concepts? I like Sam's idea of a demo or screenshare. Do you want to set one up?

text/message-tracing/README.md
swcurran: Is this HIPE proposing stack level tracing as this comment suggests?  I had assumed we were talking message receipt tracing only.  I would assume any internal tracing of the operation of the processing of a message by an agent would be handled by a separate concept and likely one that is internal to the agent - e.g. implementation specific vs. interoperable.  I don't think a DOS attack would be likely in that case.

I don't think an agent would be likely to support other than tracing of Errors in a production environment.

dhh1128:
> Would it be fine if I schedule something on Monday or maybe send an offline demo by tomorrow?

Yes, let's set something up, possibly on Tue.

TelegramSam:
How would the protocol-discovery protocol discover support for tracing? Is that valuable?

dhh1128:
> How would the protocol-discovery protocol discover support for tracing? Is that valuable?

@TelegramSam That's why I originally named the protocol the "feature discovery" protocol instead of the "protocol discovery" protocol--because I thought we might need to discover whether agents supported decorators, not just message families.

Another way to analyze tracing is to say that the decorator asks for the recipient to initiate a single-state protocol (emit a notification). Since the party interested in tracing is never going to interact with the other party directly, it simply signals its interest and waits to see if the protocol starts. It can't tell the difference between the other party supporting a protocol but not wanting to engage, and the other party not supporting the protocol at all.

This pattern manifests with tracing, but also with ACKs, with problem reports, etc. And interestingly, all of these are part of a "notifications" message family. All of them are triggered by something other than an initial message by the interested party.

dhh1128:
Superseded by https://github.com/hyperledger/aries-rfcs/pull/34

PR #61, opened 2018-11-13
propose HIPE: agent file format

dhh1128:


kdenhartog:
Looks good to me. One question that came up for me is how might an *.ap message fit inside a *.aw message. For example, if I'm transporting a message over multiple transport protocols, on some it may be possible (and easier) to just parse JSON. Where as, if I'm using IPFS, it would make sense to list this as a *.aw file rather than a *.json. However, once I've decoded the .aw file (assuming it's not just being passed as a giant string of Base64 text), would I then have a byte format of .ap inside the AW? I assumed because it was JSON string that the format would be .JSON file if it needed to be a file.

Also, as a side point, I'd like to bikeshed the naming convention of "star dot a dub" to "star dot a dubya" cause I always found it funny when I'd hear people say "W" that way.

dhh1128:
Re how an .ap message fits inside an .aw: I imagine that the .ap would be a base64-encoded attribute inside the .aw. Any .aw is going to have an .ap inside it to be useful--but it won't look like JSON when inspecting the outer wrapper.

TelegramSam:
I'm wondering if the right companion to Agent Wire is Agent Plaintext, not Application Plaintext. I of agent messages as either 'core' or 'application' levels, 'core' being standard level agent business like connection and relationship management. Agent would nicely contain both of these.

TelegramSam:
> Re how an .ap message fits inside an .aw: I imagine that the .ap would be a base64-encoded attribute inside the .aw. Any .aw is going to have an .ap inside it to be useful--but the .ap content won't look like JSON when inspecting the outer wrapper.

Won't the .ap file be the decrypted ciphertext inside the wire level format message in the .aw?

I thought the idea was to have a 1-1 mapping between wire level message format and the .aw file contents, and the agent message format (same as decrypted ciphertext in wire level message format) and the .ap file.

dhh1128:
> I'm wondering if the right companion to Agent Wire is Agent Plaintext, not Application Plaintext. I of agent messages as either 'core' or 'application' levels, 'core' being standard level agent business like connection and relationship management. Agent would nicely contain both of these.

I am not opposed to this. It has a nice consistency. We were calling messages "Application Messages" to emphasize the fact that they target application-level logic rather than low-level plumbing. I think @swcurran might have an opinion.

dhh1128:
> Won't the .ap file be the decrypted ciphertext inside the wire level format message in the .aw?

Yes.

> I thought the idea was to have a 1-1 mapping between wire level message format and the .aw file contents, and the agent message format (same as decrypted ciphertext in wire level message format) and the .ap file.

Yes.

You just said it better than I did. If that wasn't the impression that my answer gave, then let's erase my answer and use Sam's.

kdenhartog:
This makes sense. My hope would be to not see .ap files SHOULD NOT (using my RFC style writing 😄) on disk directly, therefore data at rest remains protected. I'm thinking this is in line with everyone else's thinking, but I want to double check.

dhh1128:
>.ap files SHOULD NOT be seen on disk directly

@kdenhartog I understand why you said this. Makes sense from a security perspective. However, some .ap files will be examples (as, for instance, all the files that attach to a HIPE documenting a message family). I imagine text editors will also want to implement .ap file support, and so will debuggers. The real rule is probably something like, "As a general rule, agent messages that are truly sent in A2A communication should be stored in encrypted form at rest." -- thus exempting stuff that's not actually transmitted anywhere. Or something like that.

TelegramSam:
@dhh1128 If you add the warning text about encryption, I believe this is ready for FCP. We talked about it on the Agent call, and had only that one concern.

dhh1128:
@TelegramSam I accepted @kdenhartog 's suggested paragraph about that topic.

swcurran:
> > I'm wondering if the right companion to Agent Wire is Agent Plaintext, not Application Plaintext. I of agent messages as either 'core' or 'application' levels, 'core' being standard level agent business like connection and relationship management. Agent would nicely contain both of these.
> 
> I am not opposed to this. It has a nice consistency. We were calling messages "Application Messages" to emphasize the fact that they target application-level logic rather than low-level plumbing. I think @swcurran might have an opinion.

In reading that section of the HIPE and the usage examples in the text ("This file is in Application Plaintext Format"), I think using Agent Plaintext might be better.  If this is only ever used in the Agent sub-context, I would agree on the "Application", but given the more broader visibility, I lean towards the use of "Agent".  It is subtle :-).

dhh1128:
I have updated the PR to match @swcurran's suggestion about "Agent Plaintext" instead of "Application Plaintext".

PR #65, opened 2018-12-03
Propose HIPE: Error reporting

swcurran:
A HIPE that proposes a new message type ("problem-report") for handling some categories of errors, and provides recommendations on the handling of a broad range of error categories, including the timeouts.


text/error-handling/README.md
```diff
@@ -0,0 +1,247 @@
...
+### The specification:
+
+[TODO: reconcile kabob case here and snake_case in localization HIPE.]
+
+```JSON
+{
+  "@type"            : "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/notification/1.0/problem-report",
+  "@id"              : "an identifier that can be used to discuss this error message",
+  "@thread"          : "info about the threading context in which the error occurred (if any)",
+  "@msg_catalog"     : "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/error-codes/123",
```
TelegramSam: I'm not sure msg_catalog should be an `@` attribute. Its applicability seems specific to error handling.

text/error-handling/README.md
```diff
@@ -0,0 +1,247 @@
...
+  "fix-hint-ltxt"    : { "en": "localized error-instance-specific hint of how to fix issue"},
+  "impact"           : "enum: message | thread | connection",
+  "where"            : "enum: you | me | other - enum: cloud | edge | wire | agency | ..",
+  "time-noticed"     : "<time>",
+  "tracking-uri"     : "",
+  "escalation-uri"   : ""
+}
+```
+
+### Fields
```
TelegramSam: Splitting the field list into Required Fields and Optional Fields would make the flexibility of this message type more clear.

text/error-handling/README.md
```diff
@@ -0,0 +1,247 @@
...
+
+**@id**: An identifier for this message, as described in [the message threading HIPE](https://github.com/hyperledger/indy-hipe/blob/613ed302bec4dcc62ed6fab1f3a38ce59a96ca3e/text/message-threading/README.md#message-ids). Although this decorator is not required, it is STRONGLY recommended for errors, because including it makes it possible to dialog about the error itself in a branched thread (e.g., suggest a retry, report a resolution, ask for more information). 
+
+**@thread**: A thread decorator that places the `problem-report` into a thread context. If the error was triggered in the processing of a message, then the triggering message is the head of a new thread of which the error message is the second member (`@thread.seqnum` = 1). In such cases, the `@thread.pthid` (parent thread id) here would be the `@id` of the triggering message. If the problem-report is unrelated to a message, the thread decorator is mostly redundant, as `@thread.thid` must equal `@id` and `@thread.seqnum` must be 0.
```
TelegramSam: sequence numbers are sender-specific, so the `secnum` should be 0 here.

TelegramSam: I'm surprised at using a parent thread instead of just a thread. Is this to avoid ambiguity between error replies when there is an ongoing thread, and error replies when there is no ongoing thread?

dhh1128: Agreed.

dhh1128: You are correct, and this verbiage (which I wrote) is wrong. I slipped into the mental error of thinking of seqnum as being the sequence in the thread, rather than the sequence of sends by this sender. We shouldn't use `pthid`, but rather `thid`. And `seqnum` is 0, not 1.

text/error-handling/README.md
dhh1128: It's related to localization. See the localization HIPE.

text/error-handling/README.md
```diff
@@ -0,0 +1,247 @@
...
+
+[TODO: reconcile kabob case here and snake_case in localization HIPE.]
+
+```JSON
+{
+  "@type"            : "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/notification/1.0/problem-report",
+  "@id"              : "an identifier that can be used to discuss this error message",
+  "@thread"          : "info about the threading context in which the error occurred (if any)",
+  "@msg_catalog"     : "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/error-codes/123",
+  "friendly-ltxt"    : { "en": "localized message", "code": "symbolic-name-for-error" },
```
dhh1128: I suggest changing `friendly-ltxt` to `explain_ltxt` or similar. The fact that this text is localized implies that it is human-friendly. It's not just a piece of friendly text--it's a piece of friendly text that *explains* the error.

text/error-handling/README.md
```diff
@@ -0,0 +1,247 @@
...
+
+**@thread**: A thread decorator that places the `problem-report` into a thread context. If the error was triggered in the processing of a message, then the triggering message is the head of a new thread of which the error message is the second member (`@thread.seqnum` = 1). In such cases, the `@thread.pthid` (parent thread id) here would be the `@id` of the triggering message. If the problem-report is unrelated to a message, the thread decorator is mostly redundant, as `@thread.thid` must equal `@id` and `@thread.seqnum` must be 0.
+
+**@msg_catalog** (required): a DID reference that provides a way to look up the error code in a catalog. The DID resolves to an endpoint that is combined with the DID fragment (e.g. `;spec/error-codes/123` in the above) to define a concrete URL with the error details. This is the same technique used for message family specifications.
```
dhh1128: "...used for message family specifications" --> "...used for message family specifications, and in fact could be a message family identifier, if the documentation for the message family includes codes for problem reports."

text/error-handling/README.md
swcurran: Updated

text/error-handling/README.md
swcurran: Updated

swcurran: Cleaned this up. I was also thinking that the `problem-report` message would be the first message in a separate thread, but as I wrote this, I think it's right that the `problem-report` is just the next message in the existing thread.

text/error-handling/README.md
swcurran: Updated

text/error-handling/README.md
TelegramSam: no, just splitting up the list in the documentation. No actual changes to the JSON at all.

TelegramSam:
I feel like there should be an error code field. There is a code for the human-readable error, but not necessarily a top-level attribute with a code that can be specified in a message family and matched when processing the error message.

dhh1128:
This is now superseded by Aries RFC 0035: https://github.com/hyperledger/aries-rfcs/pull/35

PR #67, opened 2018-12-12
FCP HIPE: trust ping protocol

dhh1128:

Describe how agents could ping one another to prove connectivity and security.


devin-fisher:
I think the protocol needs at least an optional nonce.  I think it should be required but those with better crypto chops could convince me that the overhead is not needed sometimes. 

[Cryptographic Nonce](https://en.wikipedia.org/wiki/Cryptographic_nonce)

devin-fisher:
I'm confused, if the `@id` is not required for the `ping` but the `@thread` is required for the pong and the `@thread` really requires an `thid` that is based on the `@id` of the ping. How could this logically work without an `@id`?

text/trust-ping/README.md
```diff
@@ -0,0 +1,119 @@
...
+The "trust" in its name comes from several features that the interaction
+gains by virtue of its use of standard agent-to-agent conventions:
+
+1. Messages should be associated with a [__message trust context__](
+https://docs.google.com/document/d/13ykeuY8sWFktvrL_3d5W2R8EKWprwD3vjVM7B4Lq5HY/edit#heading=h.4uyz6jh4ou1h)
+that allows sender and receiver to evaluate how much trust can be placed
+in the channel. For example, both sender and receiver can check whether
+messages are encrypted with suitable algorithms and keys.
+
+2. Messages may be targeted at any known agent in the other party's sovereign
```
swcurran: There are limitations on this that should be acknowledged. At minimum, the Domain Endpoint (Agency) will never have a private key for a Sender (at least as part of a pairwise DID), so it can't be pinged in this way. Likewise, if a Routing Agent is ONLY a Routing Agent, it will also not have a key for the Sender and so cannot respond.

text/trust-ping/README.md
```diff
@@ -0,0 +1,119 @@
...
+### Messages
+
+The trust ping interaction begins when `sender`
+creates a `ping` message like this:
+
+```JSON
+{
+  "@type": "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/trust_ping/1.0/ping",
+  "@id": "518be002-de8e-456e-b3d5-8fe472477a86",
+  "@timing": {
```
swcurran: I've made some comments in the Proposed HIPE about timing that might (or might not) affect this.

PR #68, opened 2018-12-12
propose HIPE: message timing decorator

dhh1128:

### DIFF
--- a/text/message-timing/README.md
+++ b/text/message-timing/README.md
@@ -72,9 +72,9 @@ convention that conflicts with these:
 * `_time`: Used for fields that identify a moment with both date and
   time precision. For example, `arrival_time` might communicate when a
   train reaches the station. The datatype of such fields is a string
-  in ISO 8601 format (_yyyy-mm-dd HH:MM:SS.xxx_) using the Gregorian
+  in ISO 8601 format (_yyyy-mm-dd HH:MM:SS.xxx..._) using the Gregorian
   calendar, and the timezone defaults to UTC. However:
-    * Precision can vary from minute to millisecond.
+    * Precision can vary from minute to microsecond or greater.
     * It is _strongly_ recommended to use the "Z" suffix to make UTC
       explicit: "2018-05-27 18:22Z"
     * If local time is needed, timezone offset notation ("2018-05-27
@@ -121,7 +121,7 @@ decorator. It offers a number of optional subfields:
 
 The meaning of these fields is:
 
-* `out_time`: The timestamp when the message was emitted. Millisecond
+* `out_time`: The timestamp when the message was emitted. At least millisecond
   precision is preferred, though second precision is acceptable.
 * `in_time`: The timestamp when the preceding message in this thread
   (the one that elicited this message as a response) was received.
--- a/text/message-timing/README.md
+++ b/text/message-timing/README.md
@@ -11,109 +11,24 @@ Explain how timing of agent messages can be communicated and constrained.
 # Motivation
 [motivation]: #motivation
 
-Many timing considerations influance asynchronous messaging delivery.
+Many timing considerations influence asynchronous messaging delivery.
 We need a standard way to talk about them.
 
 # Tutorial
 [tutorial]: #tutorial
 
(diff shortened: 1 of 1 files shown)


text/message-timing/README.md
```diff
@@ -0,0 +1,146 @@
...
+ideas, as described below. Field names like "expires" or "lastmod" are
+deprecated, because they don't say enough about what to expect from the values.
+
+Message families are free to invent their own where they have specialized
+needs (e.g., in genealogy, storing dates earlier than 1601 may be
+important and different enough to justify a set of date conventions for
+that problem domain). However, no message families should create a
+convention that conflicts with these:
+
+* `_date`: Used for fields that have only date precision,
```
swcurran: This HIPE is not an expected place to find a list of field name suffix conventions.  I recommend that this excellent list should be in a different HIPE - either general for field suffixes or specific to time-related field suffixes - and then reference the list here.

text/message-timing/README.md
```diff
@@ -0,0 +1,146 @@
...
+  Thus, a field that captures a last modified timestamp for a file, as
+  number of seconds since Jan 1, 1970 would be `lastmod_t`.
+* `_tt`: Used just like `_time` and `_t`, but for 100-nanosecond
+  intervals since Jan 1, 1601. This is the Windows FILETIME datatype.
+* `_sec` or subunits of seconds (`_milli`, `_micro`, `_nano`): Used for
+  fields that tell how long something took. For example, a field
+  describing how long a system waited before retry might be named
+  `retry_milli`. Normally, this field would be represented as an unsigned
+  positive integer.
+* `_el` tells how much time has elapsed in friendly, calendar based
```
swcurran: Suggest being a little more specific on this one - "_elapsed" - "el" is a bit too subtle.  I think we can afford the space.  "t" and "tt" might also have other names - "_epoch" and something else for "tt" - not sure what that would be.

text/message-timing/README.md
TelegramSam: Agree on excellence and need for a different HIPE.

dhh1128: Agreed. I have created PR #76 and moved the convention material there.

text/message-timing/README.md
dhh1128: Agreed on _el. Updated to _elapsed.
_t is a deliberate resonance with time_t in Posix. I added a note about it. _tt is just a double-precision _t. Although this isn't super self-explanatory, I think it's good enough.

text/message-timing/README.md
```diff
@@ -0,0 +1,146 @@
...
+  units as a string (`y` = year, `q` = quarter, `m` = month, `w` = week,
+  `d` = day, `h` = hour, `n` = minute, `s` = second), as in
+  "3w 2d 11h". Note that this datatype does not convert directly to
+  the pure time-based elapsed variant like `_sec`, because the duration
+  of months, years, and quarters is variable. 
+
+### Decorators
+
+Timing attributes of messages can be described with the `@timing`
```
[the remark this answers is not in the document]
dhh1128: I see where you're coming from, @swcurran , but I analyze this a bit differently. out_time is relevant no matter whether threads are active or not. And in_time and out_time could be relevant to a relay, even in routing (e.g., to measure how much time penalty is incurred by the relay). So I prefer to leave these in the HIPE. Decomposing the HIPE further seems like busy work. I agree that normal usage will devolve to just the set of fields that's relevant, but I don't think anything really belongs inside @thread, and I'd rather just leave everything together under @timing because I suspect we might think of other timing attributes in the future.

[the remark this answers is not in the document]
dhh1128: Perhaps part of the reason why I feel the way I do is that I don't analyze routing as a deeply separate concern from application-focused messages. Routing (in the form of the `forward` message) is just another protocol that some agents support. Today's simple imaginations of it will probably grow more complex over time, and the distinctions between routing and other stuff will seem less and less relevant.

dhh1128:
This PR is now superseded by Aries RFC 0032 (https://github.com/hyperledger/aries-rfcs/pull/32)

PR #69, opened 2018-12-12
propose HIPE: protocols explainer

dhh1128:


text/tictactoe-protocol/README.md
```diff
@@ -0,0 +1,279 @@
...
+}
+```
+
+...and so forth.
+
+Note that the order of the items in the `moves` array is NOT significant.
+The state of the game at any given point of time is fully captured by
+the moves, regardless of the order in which they were made.
+
+If a player makes an illegal move or another error occurs, the other
```
swcurran: I think (not certain) it would be more useful in the tic-tac-toe example to use an explicit message family error message ("cheater!!") rather than problem-report, and thus might not be the best example for the use of problem-report. Obviously not a big deal, but might be worth calling out the pros and cons.

It's interesting in that it kind of depends on the structure of the agent and it's handling of thread state vs. message type in the dispatching of messages.  I'm not sure as I type this if there is, or if there should b,e guidance on that in the thread HIPE.

Hope that makes sense...

text/tictactoe-protocol/README.md
```diff
@@ -0,0 +1,279 @@
...
+
+# Reference
+
+A reference implementation of the logic of a game is provided with this
+HIPE as python 3.x code. See [game.py](game.py). There is also a simple
+hand-coded AI that can play the game when plugged into an agent (see
+[ai.py](ai.py), and a set of unit tests that prove correctness (see
+[test_tictactoe.py](test_tictactoe.py).
+
+The game can be played interactively by running `python game.py`.
```
swcurran: This is a great HIPE - I really like it.

Are we nearing the point where the Message Family specification DID can be published on a ledger with an endpoint and this message family's spec published?  Put another way - is the HIPE the message family spec, or will it/should it eventually be published elsewhere in some other format?

text/tictactoe-protocol/README.md
dhh1128: We need to have a deeper conversation about specific messages versus generic ones. Let's do that offline, and report the outcome here.

text/tictactoe-protocol/README.md
dhh1128: The message family spec DID can be published any time. I guess I'll add that to my to-do list and see if we can get it taken care of. I've kind of held off because as soon as I do, the question of who holds the keys to the DID becomes more important. Any thoughts about that? Like, maybe all the maintainers of the HIPE repo hold the keys, or shamir secret sharing of maintainers, or something?

As soon as we have a DID Doc that defines a service endpoint for these message families, we need a place to host the service endpoint. Any ideas for that?

I guess we need to start working on a message family HIPE. Maybe the one I started on protocols needs to roll into it...

TelegramSam: we need instructions about how to register a DID on the ledger and update it, to allow anybody to create their own message families.

I'm wondering if we have a special location of the HIPE repo that serves as hosting for indy related message families. We use github directly, or indirectly through an automated publishing script, to serve the necessary pages.

dhh1128: I'm going to deliberately withdraw this PR. I think what we need instead is a "how to write a protocol (message family)" HIPE, with the tictactoe protocol checked in to the HIPE's folder as an example. So I'll be coming up with a replacement PR as soon as I can get to it. Until then, I'll leave the PR open so it can be read.

text/tictactoe-protocol/README.md
```diff
@@ -0,0 +1,322 @@
...
+    an announcement of the identifier and version of the message
+    family, and also enumerates error codes to be used with problem
+    reports.
+
+### States
+
+The states of each `player` in the protocol evolve according to the
+following state machine:
+
+![state machine](state-machine.png)
```
Patrik-Stas: Seems like typo in the state machine image. On state `my move`, when `send move` event occurs, it says "transition to `my move` or `wrap up`". I suppose this should be "transition to `their move` or `wrap up`"?

dhh1128: @Patrik-Stas Thanks for the catch. Now fixed.

dhh1128:
I have revised this HIPE significantly. The content about Tic Tac Toe is mostly unmodified, but there is now a main README that discusses what protocols are and how they should be defined and documented by a HIPE. The Tic Tac Toe HIPE is then presented as a hyperlinked example.

text/protocols/README.md
```diff
@@ -106,18 +105,8 @@ version](https://semver.org/#spec-item-7).
...
-evolves to 1.X+1; this is a backwards-compatible change from the protocol's
-perspective, because parties using the protocol can use either Y 1.X or
-Y 1.X+1 and remain interoperable. However, if the protocol uses a new
-feature introduced in Y version 1.X+1, then this causes the protocol's
-own version to be updated. If the usage of the new feature is optional in
-the protocol, then the dependency update is a backwards-compatible change,
-and the protocol's minor version gets updated. If the usage of the new
-feature is required, then [the protocol's major version gets updated](
-https://semver.org/#spec-item-8).
+Any change in the expectations of a message family, even if no message attributes
```
dhh1128: @TelegramSam : this is the only change in this commit that I don't feel comfortable with. I think your phrase "change in expectations" is not quite clear enough. When party A talks protocol version 1, and party B talks protocol version 2, they should not expect to be interoperable; the different major numbers advertises incompatibility. However, if it's A=1.3 and B=1.5, then they should expect interoperability except on some optional features that B apparently knows about but A does not.

So here's the way I'd state it: minor numbers in a protocol introduce features that are optional, such that the later version could be interoperable with an earlier version having the same major number, as long as the functionality used between the two parties is gracefully degraded. On the other hand, major versions say that no interoperability should be assumed.

dhh1128:
This HIPE has been converted to an Aries RFC. See https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0003-protocols/README.md.

PR #71, opened 2018-12-15
propose HIPE: decorators

dhh1128:

Describes the theory and conventions associated with A2A decorators.

### DIFF
--- a/text/decorators/README.md
+++ b/text/decorators/README.md
@@ -45,7 +45,7 @@ like this:
 
 ```JSON
 {
-  "proposed_time": "2019-12-23 17:00"
+  "proposed_time": "2019-12-23 17:00",
   "proposed_place": "at the cathedral, Barfüsserplatz, Basel",
   "comment": "Let's walk through the Christmas market."
 }
@@ -84,7 +84,7 @@ and the solution is a pair of decorators. `@id` can *decorate* **any** message:
 ```JSON
 {
   "@id": "e2987006-a18a-4544-9596-5ad0d9390c8b",
-  "proposed_time": "2019-12-23 17:00"
+  "proposed_time": "2019-12-23 17:00",
   "proposed_place": "at the cathedral, Barfüsserplatz, Basel",
   "comment": "Let's walk through the Christmas market."
 }
--- a/text/decorators/README.md
+++ b/text/decorators/README.md
@@ -1,0 +1,262 @@
+- Name: decorators
+- Author: Daniel Hardman
+- Start Date: 2018-12-14
+- HIPE PR: (leave this empty)
+
+# HIPE ????-decorators
+[summary]: #summary
+
+Explain how decorators work in agent-to-agent communication.
+
+# Motivation
+[motivation]: #motivation
+
+Certain semantic patterns manifest over and over again in communication. For
+example, all communication needs the pattern of testing the type of message
+received. The pattern of identifying a message and referencing it later is likely
+to be useful in a high percentage of all agent protocols that are ever written.
+A pattern that associates messages with debugging/tracing/timing metadata is equally
+relevant.  And so forth.
+
+We need a way to convey metadata that embodies these patterns, without complicating
+schemas, bloating core definitions, managing complicated inheritance hierarchies,
+or confusing one another. It needs to be elegant, powerful, and adaptable.
+
+# Tutorial
+[tutorial]: #tutorial
+
+A decorator is an optional chunk of JSON that conveys metadata. Decorators are not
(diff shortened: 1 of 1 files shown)


text/decorators/README.md
```diff
@@ -0,0 +1,262 @@
...
+  "agree": true,
+  "comment": "See you there!"
+}
+```
+These chunks of JSON are defined independent of any particular message schema, but
+are understood to be available in all schemas.
+
+### Basic Conventions
+
+Decorators are defined in HIPEs that document a general pattern such as [message threading](
```
swcurran: It would be nice if threads had the same specification semantics as message families - e.g. versioning, an ID and mechanism to find the specification. A HIPE seems insufficient - although I suppose we could tag the commit of the HIPE for versioning :-).

text/decorators/README.md
```diff
@@ -0,0 +1,262 @@
...
+https://github.com/hyperledger/indy-hipe/blob/7bd05ee7191d5175dd6606bb5851980076b310aa/text/message-threading/README.md)
+or [message localization](https://github.com/hyperledger/indy-hipe/pull/64).
+The documentation for a decorator explains its semantics and offers examples.
+
+Decorators are recognized by name. The name must begin with the `@` character (which
```
swcurran: Should probably require that decorators should be specified as being *explicit* - e.g. must be included in the message type spec to be used - vs. *implicit* - e.g. should be supported by all message types. Thread is an example of the latter.  I think the _l10n fields are an example of the former.

text/decorators/README.md
```diff
@@ -0,0 +1,262 @@
...
+* [**Message Thread**](http://bit.ly/2SL5kab) **or thread tree**: A participant in a
+  thread could send a message containing nothing but the `@thread` with one or
+  more additional decorators *inside* it. This should be viewed as a statement of semantics that
+  apply to subsequent communications from the sender, on that thread, until further
+  notice or until overridden at the scope of an individual message instance. (While
+  powerful, applying a decorator to this scope may be a burden on receivers because
+  it requires them to be stateful. Therefore this usage, though theoretically possible,
+  may be undesirable. If community members want to explore this, they are encouraged
+  to discuss broadly, first.)
+* **Connection**. It is conceivable that Alice could bundle up a set of decorators
```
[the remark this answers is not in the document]
dhh1128: I am not suggesting that we ever say that a *decorator **type** is viable* for a thread or connection; I am suggesting that we might want to allow someone to say that a *decorator* **instance** *is being* **applied** to a thread or connection.

text/decorators/README.md
dhh1128: Did you mean "would be nice if *decorators* had the same..."?

text/decorators/README.md
TelegramSam: I like this thought. As part of the A2A spec, we can list the specific decorators and versions that are expected to be universally used. This list can be small, and be updated over time.

text/decorators/README.md
```diff
@@ -0,0 +1,262 @@
...
+
+```JSON
+{
+  "note": "Let's have a picnic.",
+  "note.@l10n": { ... }
+}
+```
+In this example, taken from the localization pattern, `note.@l10n` decorates `note`.
+
+Besides a single message or a single field, consider the following scopes as decorator
```
TelegramSam: I don't understand this section at all. Maybe examples would help?

text/decorators/README.md
TelegramSam: This would require that all documents describing decorators would include a version history, which is a pretty easy ask.

text/decorators/README.md
swcurran: @dhh1128 - yes - not threads, decorators.  

Has the following idea been considered? It came up in conversation here. Make a decorator a YAMF (Yet Another Message Family - hey, I've used that twice now - it's a thing!).  That is, any decorator would be defined as a sub-section of the message JSON:

```
<message family fields>
"@thread" : {
       "@id": "did:sov:123456324324234;spec/threading/1.0/thread",
       <other decorator fields>
}
```

At the cost of the `@id` name/value pair we get the following benefits:

- versioning
- specification discovery
- handling consistency with Message Families ("sub-protocols", I should say)
- the potential for dispatched handling - any JSON sub part with an `@<name>` and an `@id` could passed to the agent's dispatcher for processing (or, of course, handled locally by the message type).  The message family may not even have to know about the decorator or it's purpose to be able to handle it.

Seems like that might be a worthwhile trade-off.

Your thoughts?

PR #73, opened 2018-12-18
FCP HIPE: protocol-discovery protocol

dhh1128:


text/agent-protocols/README.md
```diff
@@ -0,0 +1,198 @@
...
+### `request` Message Type
+
+An `agprot/request` message looks like this:
+
+[![request](request.png)](request.json)
+
+The `query` field is a regular expression and could be quite complex.
+Often, however, it will be used as shown here, to identify a protocol
+(message family) with just the version portion wildcarded. The regex
+must match the entire name of a protocol family, from beginning to end.
```
TelegramSam: calling this a regex feels weird since the family must match exactly at matches are not going to include a wildcard in the middle. Calling it a regex might encourage lazy developers to build very insecure matching code. How about saying it should match the prefix of the message family, with at least the family specified.

text/agent-protocols/README.md
```diff
@@ -0,0 +1,198 @@
...
+
+[![response](response.png)](response.json)
+
+The `protocols` field is a JSON object that contains zero or more keys that
+match the query. Each key is a protocol version (fully qualified message
+family identifier such as `did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/tictactoe/1.0`).
+Its value is a JSON object that enumerates the roles the responding agent
+can play in the associated protocol, and the message types it can *receive*
+(not send).
+
```
TelegramSam: I like the designation of role, but the messages feels overly complex. I like the concept that you either support a message family or you don't. Designating roles does make sense, but I feel like specifying messages invites partial family support beyond just roles.

text/agent-protocols/README.md
```diff
@@ -0,0 +1,198 @@
...
+exciting uses of agents are more dynamic and unpredictable. When
+Alice and Bob meet, they won't know in advance what features are
+supported by one another's agents. They need a way to find out.
+
+# Tutorial
+[tutorial]: #tutorial
+
+This HIPE introduces a protocol for discussing the protocols an agent
+can handle. The identifier for the message family used by this protocol is
+`agprot`, and the fully qualified URI for its definition is:
```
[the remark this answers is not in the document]
swcurran: I agree with Sam.  I definitely don't like the truncated "agprot" - we don't need to be that concise, so at minimum, please spell out both words.  I also immediately jumped to thinking about transport protocols (http, BT, etc.) when I started reading this.  Agreed that Agent to Agent messaging is a protocol, it's not what comes to mind.  I like message-support as well.

text/agent-protocols/README.md
swcurran: @TelegramSam - couldn't you specify that you want to know whether they support tictactoe of any variety?  e.g. "*/tictactoe/*"?  I think the regex format makes sense.

text/agent-protocols/README.md
swcurran: Agree with @TelegramSam. I think that is a problem with the "tictactoe" HIPE and with this one.  I don't think "roles" are formal enough in message type specs. to be formally included in the response.  If we want to add an "@role" to all messages (which I don't think we do), then it would make sense, but for now, this seems too light.

I would add that I think the example here would be wrong if roles were supported.  The tictactoe spec says that the roles are "player_x" and "player_o", so there is no role called "player".  Presumably, if this was included, the response would be `"roles" : [ "player_x", "player_o" ],`.

Having said that...  who are the producers and consumers of this data?  If we assume that at minimum, the response will include the queried message family by name (did:..../message-family/...), maybe we should assume each message family includes specific methods "i-support" and "they-support" that produce and consume (respectively) the data in the response.  That way, the each message family would have context to understand the information given by the other side in the response.

text/agent-protocols/README.md
```diff
@@ -0,0 +1,198 @@
...
+### Message Catalog
+
+If any agent wants to send [`problem-report`s](
+https://github.com/hyperledger/indy-hipe/blob/6a5e4fe2d7e14953cd8e3aed07d886176332e696/text/error-handling/README.md#the-problem-report-message-type
+) to complain about something related to `agprot` issues, it should
+ensure that [the following message catalog](catalog.json) is in scope:
+
+[![error catalog for agprot protocol](catalog.png)](agprot.catalog.json)
+
+Note that `unsupported-protocol-or-version` is NOT sent when a query produces
```
swcurran: I don't understand this statement. If there is no match for the regex, there is an empty response.  Not sure how that is different from this error.  Perhaps an example?

text/agent-protocols/README.md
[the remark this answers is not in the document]
dhh1128: I can certainly make the identifier longer.

Regarding the use of "protocol" versus "message family": I was trying to shift the community's thinking a bit, here, with my use of the word "protocol." We've been talking about message families, which is a good construct. But I think it's a bit suboptimal because it puts the focus on data and data format. As a result, I think we neglect some other issues that also need specification. For example, you've heard me harp on the need for a state machine that tells which messages can be sent, when, and how state changes as a result. I also think the *roles* in a protocol need definition.

This was, in a way, the thrust of my talk at HL Global Forum: that it's about protocols between agents. Message families are a subset (the skeleton) of a protocol. A protocol could cross message families (e.g., we could have a general ACK message that gets used in many protocols; we already have a general problem-report message that does).

@swcurran 's comment that "Agent to Agent messaging is a protocol" gives me pause, because when I used the word "protocol" here, I was thinking of *specific* subprotocols like tictactoe or credential issuance, not the general protocol infrastructure. OSI model all over again?

So: I'm pushing back about calling this "message support". It's protocol support, of which message types are a primary consideration, but not the only one. Does my pushback move the needle at all, @TelegramSam or @swcurran ?

text/agent-protocols/README.md
dhh1128: @TelegramSam Can you say more about "feels overly complex"? I'm not sure I know which complexity you're feeling uncomfortable about.

I don't think it's possible to say, "Either you support the message family or you don't." If you write an agent that is capable of receiving credentials but not issuing them, do we have to say it doesn't support the credential issuance protocol because it can only do half of it?

I introduced the complexity here because I have been thinking deeply about how to test the support that an agent claims for a protocol, and I've become convinced that we can't just declare support in general. We need to declare named roles for each protocol, and agents need to make a claim about which of them they support. In credential issuance, saying that your agent supports the protocol isn't nearly enough; we need to know whether your agent supports the protocol *as an issuer* and *as a holder*. That makes a world of difference.

Stephen is right that our current docs are too light on this topic. But we should fix the docs, not perpetuate the error.

I don't think we need @ROLE on any messages, ever. We just need to know what roles exist, and who can send which types of messages under which circumstances.

@swcurran is also right about the tictactoe example being wrong. I actually realized that before I checked in my code, but I couldn't decide whether to update tictactoe (to have one role only) or update this HIPE instead. So: which change do you think is preferable?

text/agent-protocols/README.md
dhh1128: This error isn't used in the protocol. So maybe I could just remove it. But it suddenly occurred to me that across all time, one of the most common errors that will ever be sent from one agent to another is one that complains about an unrecognized protocol or version, and we needed a place to declare it. (Imagine Alice sends a "tictactoe 2.0" message to Bob, but Bob either doesn't know about tictactoe at all, or he only knows about version 1.0 or 3.0. He has to report an error to Alice. This would be the message he'd use.) I thought this HIPE might be a logical place to declare such an error, even if the protocol doesn't use it, but maybe there's some better place...

text/agent-protocols/README.md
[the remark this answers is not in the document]
swcurran: Interesting response - I'll have to think about it more.  When I consider what you did with tictactoe, I don't think if it as a parallel concept to, say, http. That is - the word "protocol" does not come to mind (but maybe it should). Set of related messages (or a family) does feel right - which I don't think of as just data and data format - a message family includes the processing behaviour.

text/agent-protocols/README.md
swcurran: Good argument for the need for roles and agree that because the spec just talks about the interface - messages and parameters - there is not a formal place for roles. Just as there is not a formal place for expected behaviour and for state machine.  So maybe we should add more formalism to the tictactoe HIPE (or maybe I just didn't pay enough attention). For example, maybe we define some JSON for the roles and (possibly - seems hard) for the state machine that can make generating a message skeleton and tests easier.

As to the roles for tictactoe - another interesting question. My gut is that there should be just "player" and whomever it is that goes first is "x", other player is "o" - I think that's the protocol used by (most) humans.  Of course, on the rematch, that might change :-), but perhaps that's just getting unnecessarily into the weeds :-).

text/agent-protocols/README.md
swcurran: That makes sense - thanks.  I think this would be better placed into the Message Types HIPE to give it higher visibility that all Agent developers would see vs. in this HIPE which seems like just a YAMF (yet another message family).

text/agent-protocols/README.md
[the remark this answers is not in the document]
dhh1128: @swcurran A protocol is not just something like HTTP; it's any set of coordinated behaviors by multiple parties. Buying a house is a protocol. Ordering a meal at a restaurant is a protocol (and Vapianos, where we ate the other night, had a different one from many). Issuing a credential is a protocol. Issuing a credential is more than a message family; it's also all the steps that the parties have to engage in, in between sending each other messages.

text/agent-protocols/README.md
dhh1128: Okay, you convinced me. I'll put it somewhere else.

text/agent-protocols/README.md
TelegramSam: @dhh1128 Specifically, I think that specifying BOTH role and supported messages is redundant, to negative side effect. If you support role X, then you should support all messages expected to by role X, as designated in the message family spec.Given the conversation, I'm unaware of any flow where the requirements are not bet by specifying `role` alone. Why is messages necessary?

text/agent-protocols/README.md
TelegramSam: I think there are a few major use cases:
1. Tell me everything you support (use .*, but why just not specify a query?)
2. Which message families of group X do you support? (group is in this case the DID used to spec. the families)
3. Which versions of family X do you support?
4. Do you support version 1.2 of Family X

All of these are supported by prefix matching, not a general regex. Thinking of it as a prefix match makes it obvious the level of disclosure that is being requested.

I think that asking for any family that matches `/tictactoe/` would be a minor case at best.

Using a regex for this makes it much more difficult to evaluate the potential disclosure being requested, as many regexes would allow full disclosure. Full disclosure allows DIDs to be 'fingerprinted' through the families they support. I think approaching this as a regex is likely to lead to developer error to the effect of allowing full disclosure by default.

text/agent-protocols/README.md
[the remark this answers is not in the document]
TelegramSam: Good Discussion.  I think that Message Families are the _interface_ of the protocol. A reasonable assumption can be made that if you claim to support the interface, you also support the steps in between.

I think what this HIPE does is describe **discovery** of supported message family / protocol interfaces.

In regards to needle moving: I'm becoming convinced that explicitly declared 'roles' for message families and discovery of supported roles is very necessary in message family definition.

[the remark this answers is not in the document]
swcurran: Totally understand and agree on what protocols are - it's just the usage of the term in technical contexts.  Tech use of "protocol" implies putting the "thing" on a pedestal and "message families" don't feel like that.

text/agent-protocols/README.md
dhh1128: @TelegramSam If I support the "issuer" role in the "credential-issuance" protocol, that does not necessarily mean that I support negotiating on terms and conditions (something that I imagine a "credential-issuance" protocol might eventually contain). In other words, there could be fancier and simpler variants of support for a given role, even within a given version of the protocol.

It is not clear to me whether we need to be able to *discover* that fanciness, though; maybe it's enough to know that the other agent supports a role, and we can learn as we go along whether they are capable of being fancy. If not, and they return a graceful error, maybe that's good enough.

text/agent-protocols/README.md
dhh1128: @TelegramSam : I agree that a regex is unnecessarily powerful--and in fact, it imposes a requirement that agents support regexes, which is onerous for some embedded platforms. What would you think about a general wildcard, '*'. This can be implemented without any complex regex, and still provides pretty good flexibility.

text/agent-protocols/README.md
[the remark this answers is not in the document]
dhh1128: @TelegramSam and @swcurran Please go read my HIPE about how to define protocols (#69), and then see if you are still disliking my emphasis on "protocol" instead of "message family". I really think "message family" is too narrow (look at how much of the tictactoe protocol is not about the message family), and I also don't resonate to the concern that tech use of "protocol" implies putting the thing on a pedestal.

text/agent-protocols/README.md
dhh1128: Removed.

text/agent-protocols/README.md
dhh1128: I updated the spec to try to make it clear that specifying messages is often not necessary. I still think the option of specifying them makes sense, so I didn't take it out entirely, just made it optional.

text/agent-protocols/README.md
dhh1128: I took out regex and replaced just with the * wildcard.

text/feature-discovery/README.md
```diff
@@ -0,0 +1,189 @@
...
+https://github.com/hyperledger/indy-hipe/blob/4a17a845da932609f1c6b7b8a4599bb686a1f440/text/protocols/tictactoe-1.0/README.md
+).
+
+Any agent may send another agent this message type at any time.
+Implementers of agents that intend to support dynamic relationships
+and rich features are *strongly* encouraged to implement support
+for this message, as it is likely to be among the first messages
+exchanged with a stranger.
+
+##### `response` Message Type
```
TelegramSam: I'd like to suggest changing `response` to `disclose`. This better fits a flow where I volunteer features I support.

text/feature-discovery/README.md
```diff
@@ -0,0 +1,189 @@
...
+protocol could be used to mine information suitable for agent fingerprinting,
+in much the same way that browser fingerprinting works. This is antithetical
+to the ethos of our ecosystem, and represents bad behavior. Agents should
+use `feature-discovery` to answer legitimate questions, and not to build detailed
+profiles of one another. However, fingerprinting may be attempted
+anyway.
+
+For agents that want to maintain privacy, several best practices are
+recommended:
+
```
TelegramSam: I suggest as the first element of this list:

Only reveal supported features based on trust to the relationship
Even if you support a feature, you may not wish to engage it with all of your relationships. Don't tell others about features you do not plan to use with them.

text/feature-discovery/README.md
```diff
@@ -0,0 +1,189 @@
...
+# Drawbacks
+
+# Rationale and alternatives
+
+# Prior art
+
+# Unresolved questions
+
+- Do we want to support the discovery of features that are not protocol-related?
```
TelegramSam: Maybe, but that might be a different flow. What if this were called the `protocol-discovery` message family?

text/feature-discovery/README.md
```diff
@@ -0,0 +1,189 @@
...
+[![state machines](state-machines.png)](https://docs.google.com/spreadsheets/d/1smY8qhG1qqGs0NH9g2hV4b7mDqrM6MIsmNI93tor2qk/edit)
+
+### Messages
+##### `request` Message Type
+
+A `feature-discovery/request` message looks like this:
+
+[![request](request.png)](request.json)
+
+The `query` field may use the * wildcard. Usually this will be to
```
TelegramSam: Should we explicitly mention just using the * wildcard to indicate 'anything you want to share with me'?

TelegramSam:
I've added a few inline comments. Aside from those minor suggestions, my only reservation is the fairly early definition of roles in protocols. I don't want to wait though, and as it's an optional field I think we are nearly ready for FCP on this HIPE.

text/feature-discovery/README.md
dhh1128: Agreed. Will update.

text/feature-discovery/README.md
dhh1128: Agreed. Will update.

text/feature-discovery/README.md
dhh1128: Yes, I think that might be better.

text/agent-protocols/README.md
dhh1128: I removed the messages entirely.

text/feature-discovery/README.md
dhh1128: Done.

text/feature-discovery/README.md
dhh1128: Done.

text/feature-discovery/README.md
dhh1128: Done.

text/feature-discovery/README.md
dhh1128: Done.

dhh1128:
@TelegramSam I believe all outstanding issues have been resolved. The PR needs to be renamed to call it the "protocol discovery" PR instead of the "feature discovery" PR.

text/protocol-discovery/README.md
```diff
@@ -0,0 +1,190 @@
...
+and rich features are *strongly* encouraged to implement support
+for this message, as it is likely to be among the first messages
+exchanged with a stranger.
+
+##### `disclose` Message Type
+
+A `protocol-discovery/disclose` message looks like this:
+
+[![sample disclose](disclose.png)](disclose.json)
+
```
swcurran: This JSON response structure goes against the proposed [DIDComm Messaging Best Practices](https://docs.google.com/document/d/1flfXrvZ0ehzB6v69VjVLJXFyITCLjla58d7SNYVNhac/edit#heading=h.toibew5infl1)

swcurran:
Looks good!  Let's move it forward.

text/protocol-discovery/README.md
dhh1128: @swcurran Can you say more about this? I am not seeing a conflict. What specific issue is there?

dhh1128:
@TelegramSam and @swcurran I wonder if we need to provide a way to find out which decorators an agent will support? Obviously, it must support any that are required by a protocol--but would it be important to know whether it supports ~timing.expiration_time, for example? Or ~please_ack? Or...

text/protocol-discovery/README.md
dhh1128: Never mind. I think I see it. It's the guideline about "No unbounded objects". I will leave some feedback about this guideline in the doc. I don't think the guideline is framed correctly yet.

devin-fisher: I don't see why an unbounded dictionary is better than:
```
{
   "protocols":[
      {
         "protocol_type":"did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/tictactoe/1.0",
         "roles":["player"]
      }
   ]
}
```
The only thing I can think of is a quicker lookup but the recieving party can put the data into any form that facilitate its needs including lookup.

dhh1128: @devin-fisher and @swcurran I have updated this HIPE to match the conventions in the doc, now that we have clarified the whys and wherefores a bit.

TelegramSam:
> @TelegramSam and @swcurran I wonder if we need to provide a way to find out which decorators an agent will support? Obviously, it must support any that are required by a protocol--but would it be important to know whether it supports ~timing.expiration_time, for example? Or ~please_ack? Or...

Heh. I just left this comment on the message tracing proposed HIPE. I do think that is valuable, but maybe not necessary for this initial HIPE.

TelegramSam:
So, if we provide a docs link to each decorator like we do for message families, they become automatically applied to this discovery process. Seems worth discussing over in the decorator HIPE.

text/protocol-discovery/README.md
```diff
@@ -0,0 +1,190 @@
...
+exciting uses of agents are more dynamic and unpredictable. When
+Alice and Bob meet, they won't know in advance which features are
+supported by one another's agents. They need a way to find out.
+
+## Tutorial
+[tutorial]: #tutorial
+
+This HIPE introduces a protocol for discussing the protocols an agent
+can handle. The identifier for the message family used by this protocol is
+`protocol-discovery`, and the fully qualified URI for its definition is:
```
dbluhm: I think all of the other family names are using snake case versus kabob case. Switching this one to `protocol_discovery` to stay consistent seems like a good idea.

dhh1128:
Agreed. Will switch.

dhh1128:
This is superseded by Aries RFC 0031 (https://github.com/hyperledger/aries-rfcs/pull/31)

PR #78, opened 2019-01-03
propose HIPE: attachments

dhh1128:

Describes a general-purpose attachment mechanism.


TelegramSam:
I think this HIPE contains great thinking but lacks application examples necessary to evaluate its applicability as a decorator. 

I'm wondering if we need a way within the HIPE process to group 'future thinking' or 'tabled ideas' so that we don't have to lose them from the community conversation, but don't yet have to accept something that we don't yet have enough experience or information to evaluate.

kdenhartog:
> I think this HIPE contains great thinking but lacks application examples necessary to evaluate its applicability as a decorator.

I would say a perfect example of this will be with the driver's license use case for credential exchange. When the Department of Transportation wants to issue a digital driver's license they'll need to issue a picture along with the credential. In this use case, this would be a perfect decorator to use... once we convert credential exchange to an A2A message family.

swcurran:
Not sure this would be a good example for what @kdenhartog references.  It's absolutely a real use case, but the picture will be a field in the driver's licence Credential, not a decorator on a Message.

kdenhartog:
In that example, I'm making the assumption that credentials will be converted to a message family which would then make use of the decorator. Without making this assumption though, I could think of another use case.

When a bank is issuing a loan in the beginning, they'll support both digital and paper documents. The digital form will use credentials, but may also require an image of the document which has been physically signed. In this case, the message for the loan protocol would need to support a message type which includes both credential data and the image of the signed document. This idea could be extrapolated for any image which requires provenance.

Additionally, this allows for two identity owners to be certain that they are viewing the same content that exists behind a dynamic link. This is advantageous to build more complex protocols that relate to some digital content.

These are just two examples where this decorator would be useful. I'm sure there are more, leading me to believe that a general solution would be adequately used in message family designs.

kdenhartog:
This work is similar to the one hashlink work that just came through the W3C-CCG as a work item. Should we be aligning this HIPE with the hashlink spec in any way? Here's a link to the spec. https://tools.ietf.org/html/draft-sporny-hashlink-02

text/attachments/README.md
```diff
@@ -0,0 +1,278 @@
...
+recommended.
+
+* `lastmod_time`: A hint about when the content in this attachment was last
+modified.
+
+* `byte_count`: Optional, and mostly relevant when content is included by
+reference instead of by value. Tells the receiver how expensive it will be,
+in time, bandwidth, and storage, to fully fetch the attachment.
+
+* `content`: A JSON object that gives access to the actual content of the
```
tplooker: Should the base64 representation you showed in the above example be listed below?

TelegramSam:
Love the new additions.

Should inline attachments use the `~attach` attribute suffix? as in `someattribute~attach` when the value is an inline attachment? This signals that the value is able to be processed as an attachment structure.

I think the attachment block could benefit from an optional `description` attribute. If I get a message from my tax service with a variety of attachments, each could contain their own description.

Also relevant to the discussion is the forward message, which contains a message to forward.

dhh1128:
@TelegramSam Do you mean *embedded* attachments? They do use the attachment structure, whereas the *inline* attachments do not.

Agreed on the `description` attribute.

Thanks for pointing out the `forward` message. I had forgotten that. I'll look for a place to mention it.

TelegramSam:
Yes, I meant embedded.

mwherman2000:
Especially in the _appending_ scenario (e.g. https://github.com/hyperledger/indy-hipe/blob/bd48bae3712e659c34d88c6dea839ccf5a0f0701/text/attachments/README.md#appending), I think it would be use to have an `id` attribute to support the ability to directly reference at particular attachment, n'est pas ...using a mechanism similar to _DID Resolution_ (i.e. notationally but not specifically).

By extension, I think this (the `id` attribute) also makes sense for the other two scenarios: _inlining_ and _embedding_.

swcurran:
A minor whine about the use of JSON pictures instead of just using GitHub code highlighting.  Yes, the picture is (slightly) clearer, but when, for example, looking at the raw MD diffs to see what changed, we can't see the JSON at all.  Further, it's more difficult to do a subsequent PR that changes the JSON. I assume there are directions about how the pictures are generated?

My recommendation is that we stick to Github JSON code highlighting.

dhh1128:
@swcurran You have brought this up before, and I've answered it before. I'm wondering if I didn't make it clear enough, so let me try again and see if I can alleviate the concern. If not, maybe I should abandon the strategy.

The README.md references images like this:
`[alt text](x.json)`

What this means is that every image is linked to a JSON file that shows the diff. There will be line-item diffs any time the JSON in an image changes; they just won't be in the main README.

If I embed the JSON directly, then I lose the ability to use callouts, arrows, and other annotations. I'm doing that in the images here.

Did that resolve your concern, or does it still feel like it's not worth it to you?

dhh1128:
@mwherman2000 Thanks for the review! What you are asking for is what the `nickname` attribute is doing. `nickname` was chosen as the name because in email attachments, being able to refer to them in a human-friendly way is useful. We could rename it `id`... I know that's what you've become used to because of your exposure to `id` in DID Docs. However, DID Docs are schizophrenic, because JSON-LD actually uses (and mandates!) `@id` in some places, not `id`. Apparently the usage of `id` vs. `@id` in JSON-LD circles has changed over time. See point 4 in this comment: https://github.com/w3c-ccg/did-method-registry/pull/33#pullrequestreview-217908493

We are currently using JSON-LD's `@id` to identify messages, so if you really want this field name to change, my inclination would be to vote for `@id` (already permitted and defined by JSON-LD) instead of `id`. How strongly do you prefer `id`?

dhh1128:
@TelegramSam What do you think about @mwherman2000's suggestion to rename `nickname`? I don't have that strong of an opinion either way, but just want to do what several people vote for.

mwherman2000:
@dhh1128 @TelegramSam The rationale for maintaining a consistent name for the "`id`" will become clear by the end of the weekend.  I'm preparing a new specification, the _Decentralized Identifier URL Specification (`did-url-spec`)_, that separates out the `did-url` grammar in such away that the ABNF notation for the `did-url` can be some shared technology across every possible use scenario -- and not just _DID Resolution_.  To read the Purpose of the specification, click [here](https://github.com/mwherman2000/did-url-spec/blob/master/README.md#purpose).

The Indy _feature discovery_ and _attachments_ HIPEs helped to shed a very bright light on the need for this shared `did-url` specification, grammar, and parser reference implementation. ...and in doing so, I have been able to easily accommodate the "DID needs" of these 2 HIPEs as well as the following additional (primary) application scenarios:

- [W3C Decentralized Identifier specification](https://github.com/w3c-ccg/did-spec)
- DID Resolvers and [DID Resolution](https://github.com/w3c-ccg/did-resolution)
- Agent-to-Agent (A2) Communications services in the Hyperledger Indy Agent framework - for example,
   - the [Indy draft feature-discovery 1.0 HIPE](https://github.com/dhh1128/indy-hipe/blob/9c7722d208cfe0a336cb67a626cbbb192ae73f8c/text/feature-discovery/README.md), and
   - the [Indy draft attachments HIPE](https://github.com/hyperledger/indy-hipe/blob/bd48bae3712e659c34d88c6dea839ccf5a0f0701/text/attachments/README.md#hipe-00-attachments).

A draft of the `did-url-spec` presentation can be found here: https://github.com/mwherman2000/did-url-spec/tree/master/src

...but I encourage you to wait until Monday morning to watch the YoutTube webcast of this presentation.

TelegramSam:
I don't love nickname, but don't love id as it conflicts with @id. Does json-ld allow for the @id to be relative to the outside message? So...
````
{
"@type":"...",
"@id":"234567890",
"image~attach": {
  "@id": "image5",
  ...
}
}
```
And the image attachment could be referred to as `234567890/image5`?

Rough thought, but based on this from the jsonld spec: 

> JSON-LD allows IRIs to be specified in a relative form which is resolved against the document base

mwherman2000:
That might be workable too.  I'm uploading the webcast now.  Will post the URL later this afternoon (3 hour upload time).

So let me ask, should the `did-spec` also consider using `@id` instead of just `id` as the name of the "id" attribute?  

I like the composition idea a lot.  Some care/guidelines would be needed to ensure the composite identifiers are still valid "whatever" specific type of identifiers ...for example, what if the value of both `@id` attributes were both `did-url`s ...then the composition wouldn't be a valid `did-url`?

mwherman2000:
This video is available here (first in the playlist):
_Hyperonomy Universal Decentralized Identifier URI Specification (did-uri-spec) v0.14_ (https://www.youtube.com/playlist?list=PLU-rWqHm5p45c9jFftlYcr4XIWcZb0yCv).

The PPT itself can be found here: https://github.com/mwherman2000/did-uri-spec/tree/master/src

dhh1128:
I'm really liking @id as a relative rather than absolute identifier, when not at the root level of message.

dhh1128:
@TelegramSam and @mwherman2000 I have updated the HIPE to use `@id` instead of `nickname`

text/attachments/README.md
```diff
@@ -0,0 +1,439 @@
...
+
+Data can be "attached" to DIDComm messages in 3 ways:
+
+1. Inlining 
+2. Embedding
+3. Appending
+
+#### Inlining
+
+In __inlining__, data is directly assigned as the value paired with a JSON key
```
swcurran: I wonder if inlining is truly an attachment.  It's a useful way to include an attachment and so should be described, but I question the assertion that the protocol version can be maintained when the inlined JSON format has breaking changes.  There is no indicator that the field is an attachment, and it's a careful implementation that would support handling the data has an attachment.

Perhaps that assertion should be removed, or perhaps some other indicator should be added to instruct implementers that this item is an attachment that could change without bumping the protocol version.

dhh1128:
Superseded by Aries RFC 0017: https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0017-attachments/README.md

PR #79, opened 2019-01-07
DRAFT: digital signatures decorator and message family

kdenhartog:
Signed-off-by: Kyle Den Hartog <kdenhar@gmail.com>

[rendered text](https://github.com/kdenhartog/indy-hipe/tree/sigs/text/digital-signatures)

### DIFF
--- a/text/digital-signatures/README.md
+++ b/text/digital-signatures/README.md
@@ -35,7 +35,7 @@ The signature message family contains a single message format currently. This me
 {
     @type: "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/signature/1.0/ed25519Sha512_single"
     "signature": <digital signature function output>
-    "sig_data": <base64URL(64bit_integer_from_unix_epoch||message_data)>
+    "sig_data": <base64URL(formatted timestamp||message_data)>
     "signers": <signing_verkey>
 }
 ```
@@ -49,7 +49,7 @@ The `~sig` feild decorator may be used with any field of data. Its value should 
 "example_field~sig": {
     @type: "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/signature/1.0/ed25519Sha512_single"
     "signature": <digital signature function output>
-    "sig_data": <base64URL(64bit_integer_from_unix_epoch||message_data)>
+    "sig_data": <base64URL(formatted timestamp||message_data)>
     "signers": <signing_verkey>
 }
 ```
@@ -74,6 +74,22 @@ When using the signature message family or message decorator we base64 encode th
 Once the verifier has verified the data, they can base64 decode the sig_data and use the perform computations on the JSON structure. It's also worth noting that the data is not included in both encoded and decoded format in order to prevent message bloat.
 
 Another important consideration is the usage of a list for the `signers` field. Supporting a list rather than only a single string allows for multi-signature support.
+
+### Timestamp format
+
+In order to handle timestamping, it's proposed that we prepend a string representation of the ISO 8601 standard format onto the front of the message. More specifically, the format should follow `yyyymmddThhmmssZ` where,
+
+- `yyyy` -> year
+- `mm` -> month with a 0 in the first position if less than 10
+- `dd` -> day with a 0 in the first position if less than 10
+- `T` -> standard delimiter that MUST be included to indicate the characters to the right represent the time
+- `hh` -> hour using 24-hour format where a 0 is in the first position if less than 10
+- `mm` -> minutes using a 60-minute format where a 0 is in the first position if less than 10 and 00 is used if it's the beginning of the hour
+- `ss` -> seconds 
+- `Z` -> standard timezone indicator which indicates UTC timezone. The timezone must remain `UTC`, so Z should always be used.
+
+For example, `20190107T143700Z` would be the time stamp of Janurary 7th, 2019 at 2:37:00 PM UTC time zone. 
+
 
 ### Signing
 
@@ -113,6 +129,10 @@ This decorator should support a specific set of signatures while being extensibl
 
 TODO provide template in this HIPE directory
 To add a new signature scheme, follow the template which is provided to detail the new scheme as well as provide some test cases to produce and verify the signature scheme is working.
+
+### Examples
+
+TODO add example messages with this standard signature format
 
 ## Drawbacks
 [drawbacks]: #drawbacks
--- a/text/digital-signatures/README.md
+++ b/text/digital-signatures/README.md
@@ -33,11 +33,10 @@ The signature message family contains a single message format currently. This me
 
 ```
 {
-    @type: "signature/1.0/ed25519Sha512/single/"
+    @type: "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/signature/1.0/ed25519Sha512_single"
     "signature": <digital signature function output>
     "sig_data": <base64URL(64bit_integer_from_unix_epoch||message_data)>
     "signers": <signing_verkey>
-    "time_t": <64bit_integer_from_unix_epoch>
 }
 ```
 
@@ -47,12 +46,11 @@ The `~sig` feild decorator may be used with any field of data. Its value should 
 
 ```
 @type: 
-"example_feild~sig": {
-    @type: "signature/1.0/ed25519Sha512/single/"
+"example_field~sig": {
+    @type: "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/signature/1.0/ed25519Sha512_single"
     "signature": <digital signature function output>
     "sig_data": <base64URL(64bit_integer_from_unix_epoch||message_data)>
     "signers": <signing_verkey>
-    "time_t": <64bit_integer_from_unix_epoch>
 }
 ```
 
@@ -65,7 +63,6 @@ The `~sig` feild decorator may be used with any field of data. Its value should 
 2. call indy_crypto_sign(wallet_handle, signer_vk, time_and_message_data)
 3. base64URLencode the outputted data of step 2
 4. serialize data into JSON format
-5. 
 
 ### Verify Signature Implementation
 
@@ -138,8 +135,7 @@ JSON Web Signatures is an alternative to this specification. The reason we've ch
 [unresolved]: #unresolved-questions
 
 Does there need to be an signature suite agreement protocol similar to TLS cipher suites?
-
-How should we incorporate the `~timing` decorator rather than using `timestamp`?
+    - No, rather the receiver of the message can send an error response if they're unable to validate the signature.
 
 How should multiple signatures be represented?
     - One solution is to do [<digital_sig1>, <digital_sig2>] for `signature` and do [<verkey1>, <verkey2>] for `signers`
--- a/text/digital-signatures/README.md
+++ b/text/digital-signatures/README.md
@@ -4,57 +4,72 @@
 - PR: N/A
 - Jira Issue: N/A
 
-# Summary
+# Digital Signatures
+[digital-signatures]: #digital-signatures
+
+## Summary
 [summary]: #summary
 
 This HIPE is intended to highlight the ways that a non-repudiable signature could be added to a message field or message family.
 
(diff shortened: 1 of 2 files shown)


text/digital-signatures/README.md
```diff
@@ -0,0 +1,93 @@
...
+    "timestamp": <time following a standard>
+}
+
+```
+
+### Message decorator
+The `@sig` feild decorator may be used with any field of data. Its value should match the json object format of the Signature Message Family outlined below. 
+
+```JSON
+"example_feild.@sig": {
```
swcurran: Typo - feild

swcurran:
This is like the "attachments" HIPE in that it makes sense, but it would be great to have some use cases as to where this would be used.

I'm also intrigued as to using this with anoncrypt vs. using authcrypt.  Could I get the same impact (same business function) of authcrypt by using anoncrypt and adding the signature decorator on the ciphertext field?  I'm guessing not... :-)

tplooker:
Great HIPE, once this has matured further I might modify #75 to use this functionality to provide proof of ownership of a routing identifier.

text/digital-signatures/README.md
```diff
@@ -0,0 +1,93 @@
...
+***non-repudiability***: Non-repudiation refers to a situation where a statement's author cannot successfully dispute its authorship or the validity of the statment.
+
+## Examples
+
+### Signature Message Family
+The signature message family contains a single message format currently. This message family is to be used when an entire message needs to be signed. It will be structured in this way:
+
+```JSON
+{
+    @type: signature/1.0/ecdsa
```
tplooker: Is there some sort of standardized signature qualification we could use here?

kdenhartog:
> This is like the "attachments" HIPE in that it makes sense, but it would be great to have some use cases as to where this would be used.
> 
> I'm also intrigued as to using this with anoncrypt vs. using authcrypt. Could I get the same impact (same business function) of authcrypt by using anoncrypt and adding the signature decorator on the ciphertext field? I'm guessing not... :-)

I agree, I think a perfect first use case is with @tplooker PR #75 and we'll eventually need it for credentials as well.

With regards to the anoncrypt w/ signature decorator there is a slight difference. The difference being that with Authcrypt, repudiable authentication is used, where as with the suggestion you included it's non-repudiable authentication. These properties have subtle difference that will affect privacy which is why we've defaulted to using authcrypt with the option of adding non-repudiable signatures for authentication.

text/digital-signatures/README.md
```diff
@@ -0,0 +1,93 @@
...
+}
+
+```
+
+### Message decorator
+The `@sig` feild decorator may be used with any field of data. Its value should match the json object format of the Signature Message Family outlined below. 
+
+```JSON
+"example_feild.@sig": {
+    @type: signature/1.0/ecdsa
```
TelegramSam: is the type needed inside the field block to allow signatures different from `ecdsa`? If not, this feels unnecessary.

text/digital-signatures/README.md
```diff
@@ -0,0 +1,93 @@
...
+    "scheme": <digital signature scheme used>
+    "signature": <digital signature function output>
+    "sig_data": <base64(message_data)>
```
TelegramSam: should `message_data` be changed to `example_field` in this line?

TelegramSam:
> have some use cases as to where this would be used.

Connection protocol (and later maintenance) needs this.

TelegramSam:
As I see it, these signatures can be processed by a receiving agent prior to presenting the data to a message handler. Then the handler can access the verified signature metadata and the signed data easily.

as an alternative, convenience functions can be used to access signed data inside the handler, to simplify the processing of these signatures.

Some quick text describing that intent could be helpful in the HIPE to help explain how the complexity of using these signatures can be mitigated.

text/digital-signatures/README.md
kdenhartog: yeah this is to support any arbitrary signature scheme. For example, with bitcoin's curve we'd use ECDSA where as with ed25519 keys we'd use edDSA. This allows us to support multiple types of signatures, and we can add new signatures to the message family as we want to support them. at the beginning I'd suggest adding support only for BLS and edDSA because those are the only keys we use currently.

text/digital-signatures/README.md
kdenhartog: yeah, I believe the DID spec specifies them. This will need to be added before the propose phase.

text/digital-signatures/README.md
kdenhartog: thanks, I'll get this changed on my next update.

text/digital-signatures/README.md
kdenhartog: yes, message_data was intended to be the value that would be placed in the value section if a signature decorator isn't used. Does that align with what you're thinking?

text/digital-signatures/README.md
```diff
@@ -0,0 +1,93 @@
...
+# Prior art
+[prior-art]: #prior-art
+
+[IETF RFC 7515 (JSON Web Signatures)](https://tools.ietf.org/html/draft-ietf-jose-json-web-signature-41)
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+Does there need to be an signature suite agreement protocol similar to TLS cipher suites?
+
```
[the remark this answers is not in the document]
kdenhartog: The direction I wanted to head with this was what if the verifier doesn't support a particular signature scheme. I agree that there's no agreement necessary, but the verifier should be able to specify to the sender somehow that they don't support a particular scheme, or to specify which schemes they do support.

text/digital-signatures/README.md
```diff
@@ -0,0 +1,122 @@
...
+    "timestamp": <time following a standard>
+}
+
+```
+
+### Message decorator
+The `@sig` field decorator may be used with any field of data. Its value should match the json object format of the Signature Message Family outlined below. 
+
+```JSON
+"example_field@sig": {
```
tplooker: Is there the potential to have a signature for the entire message ie 
`
{
  "@type" : "message-type",
  "attr1" : "data1",
  "attr2" : "data2",
  "@sig" : {
    "@type": "signature/1.0/ecdsa",
    "scheme": <digital signature scheme used>,
    "signature": <digital signature function output>,
    "field_data": <base64(field_data)>,
    "signers": [list of signer's keys],
    "timestamp": <time following a standard>,
  }
}
`

text/digital-signatures/README.md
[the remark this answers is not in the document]
mikelodder7: Hmmm, that seems like a can of worms. I don't think the verifier should be able to tell the signer which keys to use all the time but it makes sense like the government can specify it only accepts certain keys.  I think  that should be a separate HIPE

text/digital-signatures/README.md
```diff
@@ -90,33 +105,41 @@ When a signed message is received, the signature should be verified and the mess
...
-
-How should we incorporate the `@timing` decorator rather than using `timestamp`?
-
-Should we be concatenating the `timestamp` data to the message_data as a part of the base64encoded data?
+How should we incorporate the `~timing` decorator rather than using `timestamp`?
 
-How should errors in signature validation be handled? Should processing be stopped and an error_report sent back to sender, or should the error be passed along in the message context for the family to handle?
\ No newline at end of file
+How should multiple signatures be represented?
+    - One solution is to do [<digital_sig1>, <digital_sig2>] for `signature` and do [<verkey1>, <verkey2>] for `signers`
```
mikelodder7: Cool idea but should be a separate HIPE.

text/digital-signatures/README.md
```diff
@@ -0,0 +1,141 @@
...
+### Signature Message Family
+[message-family]: #message-family
+The signature message family contains a single message format currently. This message family is to be used when an entire message needs to be signed. It will be structured in this way:
+
+```
+{
+    @type: "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/signature/1.0/ed25519Sha512_single"
+    "signature": <digital signature function output>
+    "sig_data": <base64URL(64bit_integer_from_unix_epoch||message_data)>
+    "signers": <signing_verkey>
```
TelegramSam: should this be `signer` insted of `signers` given that there will only be one for this signature type?

cywolf:
I could use some detail on the timestamp representation, "append timestamp as 8 bytes to the front of the message data". Is that a big-endian unsigned (long) int?

text/digital-signatures/README.md
```diff
@@ -0,0 +1,141 @@
...
+[drawbacks]: #drawbacks
+
+Since digital signature in this HIPE are non-repudiable, it's worth noting the privacy implications of using this functionality. In the event that a signer has chosen to share a message using a non-repudiable signature, they forgo the ability to prevent the verifier from sharing this signature on to other parties. This has potentially negative implications with regards to consent and privacy. 
+
+**Therefore, this functionality should only be used if non-repudiable digital signatures are absolutely necessary.**
+
+## Rationale and alternatives
+[alternatives]: #alternatives
+
+JSON Web Signatures is an alternative to this specification. The reason we've chosen to diverge from this specification is do to the inability to support our need for message decorators. Additional concerns around non-support for BLS signatures and other multi-signature schemes caused us to want to diverge.
```
swcurran: Could you explain why JWS precludes message decorators?  I don't get that from skimming the JWS spec.

The "non-support for BLS signatures" is a crypto thing beyond me, but it seems to me that the reasons need to be good for not to go with an existing spec.

kdenhartog: I believe JWS formats won't work with CL Signature schemes, so if we intend to use this decorator with credentials, it wouldn't make sense to use it. However if we're looking to do just basic ed25519 signatures JWS would likely be the best format.

mikelodder7: We could support BLS signatures for both threshold and multisigs, or even a single signer because those signatures are 1/2 the size of ECC sigs. CL sigs wouldn't work for JWS.

text/digital-signatures/README.md
mikelodder7: signers makes sense so its future proof. I vote for signers.

mikelodder7:
> I could use some detail on the timestamp representation, "append timestamp as 8 bytes to the front of the message data". Is that a big-endian unsigned (long) int?

That's what I thought it was. I agree it should be clarified to state "This timestamp field is a 64-bit unsigned big-endian integer. It records the number of seconds elapsed between January 1, 1970 UTC and the time the token was created."

kdenhartog:
> > I could use some detail on the timestamp representation, "append timestamp as 8 bytes to the front of the message data". Is that a big-endian unsigned (long) int?
> 
> That's what I thought it was. I agree it should be clarified to state "This timestamp field is a 64-bit unsigned big-endian integer. It records the number of seconds elapsed between January 1, 1970 UTC and the time the token was created."

Thank you for clarifying. I hadn't gotten around to looking up what was being used in the standard. I'll add this detail to the HIPE.

text/digital-signatures/README.md
kdenhartog: Are you seeing multi-sig format and single signer formats being the same @mikelodder7? That's what I was intending as well in which signers would make sense to use for this.

text/digital-signatures/README.md
kdenhartog: In the case of CL signatures and other considered Credential signatures would it make sense to use this format anyways? If it doesn't I see value in moving this format to align with the JWS work to allow for us to integrate with other identity systems more easily. (E.g. OAuth and openID connect servers)

kdenhartog:
What's our consensus on moving this towards JWS? I'm in favor at this point.

dhh1128:
@kdenhartog Would you agree that this should now be moved over to aries-rfcs?

kdenhartog:
I'd actually like to deprecate and close it (supersede in favor of 0066 is fine as well). This moves in a different direction than 0066 does, and I'm not wanting to separate out the decorator portion from the cryptographic envelope portion.

This is superseded by [Aries RFC #66](https://github.com/hyperledger/aries-rfcs/pull/66)

PR #80, opened 2019-01-15
FCP HIPE Update: minor tweak to introduce "Native Object" in File Format HIPE

dhh1128:


TelegramSam:
@dhh1128, I'm curious why you felt the need to add this?

dhh1128:
@TelegramSam In internal Evernym discussions, we were using the term "Agent Message" in a number of design docs, and also in informal conversations, and we were confusing one another. When we dug into it, we realized that in many cases we intended the concept of a Native Object format for a message, but lacked a word for it. Once we created a term for the concept, a lot of the confusion evaporated. So I'm just suggesting that this concept needs to be formally described and labeled. I'm not wedded to the label, and I'm also not wedded to placing the description in this HIPE (which is, after all, about file formats, NOT about Native Object format). But this was the best home for it that I could think of.

PR #81, opened 2019-01-16
FCP: BasicMessage HIPE

TelegramSam:


kdenhartog:
I think it would be useful to declare this a message type (one instance of a message family) so that we can use the HIPE update process to add the out of scope message types over time.

TelegramSam:
> I think it would be useful to declare this a message type (one instance of a message family) so that we can use the HIPE update process to add the out of scope message types over time.
> 
> outside of that, I think this works, is clean, and gets the job done. Let's FCP this before scope creep comes in. 😎

I think it should be a whole new message family for full-featured solutions. I think the full-featured families are likely to change more rapidly and having just a simple message family be independent of those will serve as a fallback option if needed.

dbluhm:
> What's the advantage of having separate message family rather than one giant message family for generalized text-based messaging?

My personal thinking (to this point at least) is that message families provide a nice grouping of messages that can be easily used to communicate the capabilities of any given agent. However, I don't think this means that we can't communicate partial support for message types within a message family. At the same time, I think a separation between the BasicMessage and other more fully featured messaging protocols would be cleaner.

dhh1128:
I vote to merge this HIPE. However, I'd like to suggest one simple change: can we call this "PrimitiveMessage" instead of "BasicMessage? This name seems more in harmony with the bare-bones target we are striving for, whereas BasicMessage sounds like something we would extend.

Even if this suggestion is not accepted, I am still okay with merging the HIPE.

PR #82, opened 2019-01-21
Initial admin messages HIPE

TelegramSam:

### DIFF
--- a/text/admin-messages/README.md
+++ b/text/admin-messages/README.md
@@ -39,6 +39,8 @@ The term Admin Console refers to software used by a human to perform remote admi
 # Examples
 [examples]: #examples
 
+Below are some examples of what admin messages would be useful. This HIPE is not specifically proposing these admin message families, but includes them as an illustration of which admin message types could be useful. If created, these Admin Message Families would need to be defined in another HIPE.
+
 **Family: [trust_ping](https://github.com/hyperledger/indy-hipe/blob/68073995bd472f1bc95259ca5a2e269b912bcc5f/text/trust-ping/README.md)** 
 
 Admin Family: trust_ping_admin
--- a/text/admin-messages/README.md
+++ b/text/admin-messages/README.md
@@ -1,0 +1,92 @@
+- Name: admin-messages
+- Author: Sam Curren <sam@sovrin.org>
+- Start Date: 2019-01-19
+- PR: (leave this empty)
+- Jira Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Admin Messages are used for agent control and administration, not for communication with agents in other domains.
+
+# Motivation
+[motivation]: #motivation
+
+Most cloud agents run headless. Admin messages are used by authorized agents to control the cloud agent. This ability is needed in the following situations:
+
+- User administration of a headless cloud agent via an Admin Console.
+- Controlling an enterprise cloud agent from a static client on a backend system.
+- Scripting Cloud Agents during Demos.
+- Triggering behavior and arranging state during testing.
+
+Standardizing admin messages also allows for tool reuse in all these situations.
+
+# Tutorial
+[tutorial]: #tutorial
+
+Admin messages are grouped into message families just like regular messages. 
+
+An admin message family usually relates to a regular message family. By convention, the name of the admin message family should use the same name as the main message family, with the suffix of `_admin`. For example, the `connections` message family should have admin messages grouped into the `connections_admin` message family. Using the `_admin` suffix will make it easier to correctly apply permissions.
+
+#### Permissions
+
+Admin family use should be limited to those agents (full or static) that have been explicitly given permission to perform administrative functions. For many cases, those keys and allowed permissions will be specified in configuration or at run time.
+
+#### Admin Console
+
+The term Admin Console refers to software used by a human to perform remote administrative tasks on a headless cloud agent. The Admin Console functions as a static agent. Initial versions are browser based, but no restrictions prevent native apps from being created, or native apps adopting admin console behavior.
+
+# Examples
+[examples]: #examples
+
+**Family: [trust_ping](https://github.com/hyperledger/indy-hipe/blob/68073995bd472f1bc95259ca5a2e269b912bcc5f/text/trust-ping/README.md)** 
+
+Admin Family: trust_ping_admin
+
+- send_trust_ping - instructs the agent to initiate a trust ping
+
+Commentary: The new send_trust_ping admin messages allows for the following use cases:
+
+- Agent Test Suite requests agent under test to initiate the trust ping in the sender role. Without this admin message, testing the agent's ability to initiate a trust ping (instead of just respond) would be impossible.
+- User requests trust ping from admin console.
+- Enterprise Cloud agent instructed to initiate trust_ping by enterprise backend system.
+
+**Family: [Connections](https://github.com/hyperledger/indy-hipe/blob/9e0d5804118235ba73948b0f866ddda70026e21f/text/connection-protocol/README.md)**
+
+Admin Family: connections_admin
+
+- connection_list_request - request a list of connections
+- connection_list - lists connections
+- send_invite - instruct agent to send invite
+- invite_sent - notice that invite was sent
+- invite_received - notice that an invite was received
+- send_request - instruct agent to send request
+- request_sent - notice that a request was sent
+- request_received - notice that a request was received
+- send_response - instruct agent to send response
+- response_sent - notice that a response was sent
+- response_received - notice that a response was received
+
+Commentary: There are more admin messages (11) than there are regular messages (3). This may be necessary due to the inherent complexity of protocol state.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Potential security hole. By standardizing and being explicit, we have a better chance of not making mistakes.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Administration via some other protocol. Now we have _two_ protocols.
+
+# Prior art
+[prior-art]: #prior-art
+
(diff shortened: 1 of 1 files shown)


text/admin-messages/README.md
```diff
@@ -0,0 +1,92 @@
...
+- send_response - instruct agent to send response
+- response_sent - notice that a response was sent
+- response_received - notice that a response was received
+
+Commentary: There are more admin messages (11) than there are regular messages (3). This may be necessary due to the inherent complexity of protocol state.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Potential security hole. By standardizing and being explicit, we have a better chance of not making mistakes.
```
blhagadorn: I think these message families (or at least the data contained within them) will either come about in an organic form through someone's agency architecture that one might create or via these "standard and explicit" methods, and I'd prefer the latter for security's sake.  As long as they are configurable and explicitly turned to "on" by the developer implementing them, I can't see any security issue.  
However, one important consideration would be that someone _forgets_ to turn off the ability to see the connection_list or other important features if someone is copying and pasting a config file, so it might be helpful to not only enforce these as automatically "off" in the config file, but also automatically off in the agent initialization of each instance.

text/admin-messages/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+Most cloud agents run headless. Admin messages are used by authorized agents to control the cloud agent. This ability is needed in the following situations:
+
+- User administration of a headless cloud agent via an Admin Console.
+- Controlling an enterprise cloud agent from a static client on a backend system.
+- Scripting Cloud Agents during Demos.
+- Triggering behavior and arranging state during testing.
```
blhagadorn: I see the immense benefit of these message families for development testing as well as regular production testing, but I think we should discern between the two as development testing might want some extreme testing that we might not ever want as a capability (super, super powers) when an agent goes into production.  Do you think defining the different testing types would be an efficient separation of concerns?
Here are some I can foresee:

1. Development testing (super powers that makes development of the SDK and agents quick and efficient without the use of mundane log files or debugging messes)
2. Demo testing (stuff that would be great for illustrative purposes in a demo and to progress further, yet again would never want to be in a production implementation)
3. Production testing (stuff that might be helpful for updates of agents in production, connectivity across agents and agencies, and other stuff).
Feel free to squash these ideas or refine this list, but I think for security as well as the mental separation of concerns, this nuanced view might help.

text/admin-messages/README.md
```diff
@@ -0,0 +1,92 @@
...
+- invite_sent - notice that invite was sent
+- invite_received - notice that an invite was received
+- send_request - instruct agent to send request
+- request_sent - notice that a request was sent
+- request_received - notice that a request was received
+- send_response - instruct agent to send response
+- response_sent - notice that a response was sent
+- response_received - notice that a response was received
+
+Commentary: There are more admin messages (11) than there are regular messages (3). This may be necessary due to the inherent complexity of protocol state.
```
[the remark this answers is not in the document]
TelegramSam: Using threaded ACK messages would certainly eliminate a pile of simpler admin messages.

text/admin-messages/README.md
TelegramSam: This is a great outline. I see this more of a permissions issue surrounding admin messages, but we might identify admin messages that should be explicitly enabled as you suggest.

TelegramSam:
@blhagadorn Not ALL these admin messages can be converted to ACKs, just some of them.

The specific messages mentioned in this HIPE were intended as examples of what types of Admin Messages would be useful. If created, Admin Messages for those Message Families would need to be defined somewhere else.

blhagadorn:
Hi all, @dbluhm drew a diagram for me earlier on a whiteboard that I think could help frame understanding and improvement of Admin Message Families.  Definitely looking for feedback and hope to make a few with more flow around the arrows for BasicMessage and TrustPing examples
image

text/admin-messages/README.md
dhh1128: I agree with @blhagadorn

text/admin-messages/README.md
```diff
@@ -0,0 +1,94 @@
...
+# Examples
+[examples]: #examples
+
+Below are some examples of what admin messages would be useful. This HIPE is not specifically proposing these admin message families, but includes them as an illustration of which admin message types could be useful. If created, these Admin Message Families would need to be defined in another HIPE.
+
+**Family: [trust_ping](https://github.com/hyperledger/indy-hipe/blob/68073995bd472f1bc95259ca5a2e269b912bcc5f/text/trust-ping/README.md)** 
+
+Admin Family: trust_ping_admin
+
+- send_trust_ping - instructs the agent to initiate a trust ping
```
dhh1128: I think there ought to be a generic admin message called "admin_start" that tells the agent to take a particular role in a protocol and make the first move. It should tell the agent the role that we want it to take, and the other party we want it to engage with. For example, { "@type": ".../1.0/start", "protocol": "...trust_ping/1.0", "role": "pinger", "other_roles": {"did:sov:otherparty": "pingee" }. Then a message that starts the basic_message protocol is just a variation. We don't need a special admin message type for every situation.

text/admin-messages/README.md
```diff
@@ -0,0 +1,94 @@
...
+- User requests trust ping from admin console.
+- Enterprise Cloud agent instructed to initiate trust_ping by enterprise backend system.
+
+**Family: [Connections](https://github.com/hyperledger/indy-hipe/blob/9e0d5804118235ba73948b0f866ddda70026e21f/text/connection-protocol/README.md)**
+
+Admin Family: connections_admin
+
+- connection_list_request - request a list of connections
+- connection_list - lists connections
+- send_invite - instruct agent to send invite
```
dhh1128: send_invite is just the admin_start message I described in my previous message.

text/admin-messages/README.md
```diff
@@ -0,0 +1,94 @@
...
+- invite_sent - notice that invite was sent
```
dhh1128: I don't understand why admin messages are needed to communicate events; shouldn't the agent notice the events because it gets an ACK or a response or something?

If you really need this, then there should be an "admin_event" message that specifies the event name (defined in the state machine for each protocol), the relevant protocol, etc. We don't need a specialized message for each one.

dbluhm: These messages are *not* to notify an agent of an event. These messages are used to notify an admin interface which interacts with the agent.

dbluhm:
> Hi all, @dbluhm briliiantly drew a diagram for me earlier on a whiteboard (hopefully I got it right) that I think could help frame understanding and improvement of Admin Message Families. Definitely looking for feedback and hope to make a few with more flow around the arrows for BasicMessage and TrustPing examples and for those that might be visually inclined.

@blhagadorn as a side note, we're pretty close to implementing a JS version of pack and unpack to encrypt and authorize even the admin messages. Just noticed in your diagram that whether the message was encrypted was a distinguishing characteristic of the kind of message but hopefully everything will be encrypted soon :slightly_smiling_face:

dbluhm:
> After reading through this, I'm thinking that we may be prematurely standardizing. My thinking is that this architecture pattern will naturally emerge and to jump on standardizing this message family too early will pigeonhole us.
> 
> I realize that this will affect your implementation of the Python agent, so my suggestion is that we place this on hold until we've had time to see how this pattern overlaps with a few other primary/replica patterns. In the meantime, go ahead and implement something for your use case and we can use your implementation and learnings to guide the standardization later.

@kdenhartog I think it's reasonable for us to continue using this concept without standardizing the messages and the format if that is ultimately what is needed; that being said, didn't we include versioning in the message types so we could iterate on our ideas? I don't think the question here is whether we're in danger of prematurely standardizing or pigeon-holing ourselves.

text/admin-messages/README.md
```diff
@@ -0,0 +1,94 @@
...
+Admin family use should be limited to those agents (full or static) that have been explicitly given permission to perform administrative functions. For many cases, those keys and allowed permissions will be specified in configuration or at run time.
+
+#### Admin Console
+
+The term Admin Console refers to software used by a human to perform remote administrative tasks on a headless cloud agent. The Admin Console functions as a static agent. Initial versions are browser based, but no restrictions prevent native apps from being created, or native apps adopting admin console behavior.
+
+# Examples
+[examples]: #examples
+
+Below are some examples of what admin messages would be useful. This HIPE is not specifically proposing these admin message families, but includes them as an illustration of which admin message types could be useful. If created, these Admin Message Families would need to be defined in another HIPE.
```
swcurran: Our thought in planning our (Indy Cat/BC Gov) agent was that a message family would define a set of decision points where it would make sense that external "business rules" could be applied to make a decision about how to proceed.  Such a "business rules engine" might be:

- a person via a UI ("Hey, you've been asked by ACME for a proof request and we found these credentials that could be used.  Please pick the ones to use."), 
- a set of preferences defined by a person that allows their agent to handle some responses automatically
- an Enterprise System that calculates a response based on data and rules,
- an Enterprise System that puts the request into a processing queue for processing (e.g. a client management system)

If that is seen as a reasonable model, then it probably makes sense for the Message Family to define the decision points so that an agent can put in a mechanism for managing subscriptions to those decision points. That would eliminate the need for naming the points/examples here and defines where they need to be added.  They would be needed in the Message Family definition - as spec'd in @dhh1128's [Protocol Explainer HIPE](https://github.com/hyperledger/indy-hipe/pull/69).

A second "agent-controlling" need for Admin messages is to initiate a protocol. As @dhh1128 has said, that can probably be handled generically by a "start-protocol" message that takes as a first argument the protocol (message family type) to be started.

TelegramSam:
I'm closing this pull request. If I bring up the concepts again, it will be in Aries RFCs.

PR #83, opened 2019-01-25
propose HIPE: JSON-LD compatibility

dhh1128:

### DIFF
--- a/text/json-ld-compatibility/README.md
+++ b/text/json-ld-compatibility/README.md
@@ -1,0 +1,256 @@
+- Name: json-ld-compatibility
+- Author: Daniel Hardman <daniel.hardman@gmail.com>
+- Start Date: 2019-01-23
+- PR: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Explains the goals of Indy agent-to-agent messaging with respect to JSON-LD,
+and how Indy proposes to accomplish them.
+
+# Motivation
+[motivation]: #motivation
+
+JSON-LD is a familiar body of conventions that enriches the expressive power of
+plain JSON. It is natural for people who arrive in the Indy agent-to-agent (A2A)
+ecosystem to wonder whether we are using JSON-LD--and if so, how. We need a
+coherent answer that clarifies our intentions and that keeps us true to those
+intentions as the ecosystem evolves.
+
+# Tutorial
+[tutorial]: #tutorial
+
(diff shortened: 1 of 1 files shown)


tplooker:
@dhh1128 I think as a community we should chat about what we formally rename the A2A protocol to, personally i think `Didcomms` is too broad as there could be a variety of different ways two dids could communicate, my preference would be to qualify it further to something like `DidMessaging` or an even  less general term

text/json-ld-compatibility/README.md
```diff
@@ -0,0 +1,256 @@
...
+ecosystem to wonder whether we are using JSON-LD--and if so, how. We need a
+coherent answer that clarifies our intentions and that keeps us true to those
+intentions as the ecosystem evolves.
+
+# Tutorial
+[tutorial]: #tutorial
+
+The [JSON-LD spec](https://w3c.github.io/json-ld-syntax/) is a recommendation
+and work product of the [W3C Credentials Community Group](https://github.com/w3c-ccg/community).
+It is not a formally approved W3C standard, but it has significant gravitas in
```
brentzundel: JSON-LD 1.0 is a published recommendation of the W3C, this means that it is a formally approved standard.

text/json-ld-compatibility/README.md
```diff
@@ -0,0 +1,256 @@
...
+plain JSON. It is natural for people who arrive in the DID Communication (DIDComm)
+ecosystem to wonder whether we are using JSON-LD--and if so, how. We need a
+coherent answer that clarifies our intentions and that keeps us true to those
+intentions as the ecosystem evolves.
+
+# Tutorial
+[tutorial]: #tutorial
+
+The [JSON-LD spec](https://w3c.github.io/json-ld-syntax/) is a recommendation
+and work product of the [W3C Credentials Community Group](https://github.com/w3c-ccg/community).
```
brentzundel: The JSON-LD standard was published by the RDF working group: https://www.w3.org/2011/rdf-wg/wiki/Main_Page
An updated spec (https://json-ld.org/spec/latest/json-ld/) is currently being developed by the JSON-LD community group.

dhh1128: Thank you for the improved links and the additional status information. I have updated the discussion to reflect that.

text/json-ld-compatibility/README.md
dhh1128: Agreed. HIPE updated.

PR #86, opened 2019-01-31
FCP HIPE: Agents explainer

dhh1128:


text/0002-agents/README.md
```diff
@@ -0,0 +1,408 @@
...
+not an agent. It doesn't represent a fiduciary or hold keys. (However,
+a fancy IoT doorbell that reports to Alice's mobile agent using an
+A2A protocol _would_ be an agent.)
+* A microservice run by AcmeCorp to integrate with its vendors is
+not an agent for Acme's vendors. Depending on whether it holds
+keys and speaks A2A protocols, it may or may not be an agent
+for Acme.
+* A human delegate who proves empowerment through keys might be
+thought of as an agent.
+* The keys for an agent can be stored on paper. This storage
```
TelegramSam: paper key storage feels more like a wallet, though different in a way than described above.

TelegramSam:
Daniel, this is excellent high level work. I've added a few comments, but overall this is excellent.

mwherman2000:
1. Checkout this taxonomy chart: https://hyperonomy.com/2019/02/01/architecture-driven-taxonomy-for-ssi-agents/ (You already have)
2. IoT things are mentioned but non-fungible entities (NFEs) aka Things are not ...can be implication in most readers minds, they are excluded.  Please mention Things or NFEs wherever IoT things are mentioned.
3. [[Essential Characteristics](https://github.com/hyperledger/indy-hipe/blob/8388429d437f99378b7d72aff9a36d78aa9db5c0/text/0002-agents/README.md#essential-characteristics)] In the Lightweight Agent scenario (https://hyperonomy.files.wordpress.com/2019/02/hbb-ssi-agents-v0.3.png), bullet 2 doesn't apply. #sorry
4. [[Canonical Examples](https://github.com/hyperledger/indy-hipe/blob/8388429d437f99378b7d72aff9a36d78aa9db5c0/text/0002-agents/README.md#canonical-examples)] Can you give each of these scenarios a name?  e.g. in order, Full Function Agent, Relay Agent, Advanced Function Agent. 
5. [[How Agents Talk](https://github.com/hyperledger/indy-hipe/blob/8388429d437f99378b7d72aff9a36d78aa9db5c0/text/0002-agents/README.md#how-agents-talk)] I know the goal is to stay somewhat high-level but my first question when I read this section is how does one _discover_ how to setup and initiate a conversion?  A sentence is needed in this section. [The steps are described near the end of the [General Patterns](https://github.com/hyperledger/indy-hipe/blob/8388429d437f99378b7d72aff9a36d78aa9db5c0/text/0002-agents/README.md#general-patterns) section ...we just need a sentence earlier about discoverability.]
6. [[General Patterns](https://github.com/hyperledger/indy-hipe/blob/8388429d437f99378b7d72aff9a36d78aa9db5c0/text/0002-agents/README.md#general-patterns)] Please add subsections for each pattern so that github will create hyperlinks for each pattern.
7. Change "ralized" to "realized" in https://github.com/hyperledger/indy-hipe/blob/8388429d437f99378b7d72aff9a36d78aa9db5c0/text/0002-agents/README.md#by-location
8. Change "has to talk" to "have to talk" in https://github.com/hyperledger/indy-hipe/blob/8388429d437f99378b7d72aff9a36d78aa9db5c0/text/0002-agents/README.md#the-agent-ness-continuum
9. and "speaks" to "talks" in the same section
10.

text/0002-agents/README.md
```diff
@@ -0,0 +1,408 @@
...
+security and privacy toolset, their links to the [DIDs and DID Docs of
+identity owners](https://w3c-ccg.github.io/did-spec/), and the ways
+their messages are packaged and handled are standard. That's what makes
+agents interoperable.
+
+Because agents speak so many different ways, and because many of them
+won't have a permanent, accessible point of presence on the network, they
+can't all be thought of as web servers with a Swagger-compatible API
+for request-response. The analog to an API construct in agent-land is
+_protocols_. These are patterns for stateful interactions. They specify
```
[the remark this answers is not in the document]
dhh1128: I like this suggestion and I am agreeing with it, but I'm struggling to find a graceful way to modify the text. Do you have any suggestions, @TelegramSam ?

text/0002-agents/README.md
dhh1128: Agreed. I've updated the verbiage to clarify this.

dhh1128:
@mwherman2000 : Thanks for the detailed comment. I have addressed the items as follows:

2: Agreed. Adjusted verbiage. I am not using the term "NFE", but I have used "thing" in its broader sense and given examples, and I have stopped being so IoT-centric.
3: You can't participate in the agent ecosystem without keys. It is a universal requirement. You *can* hold a key without a wallet (e.g., in a config file). This is how static agents were defined. You can't get so lightweight that you don't use keys. An entity that doesn't hold keys can call an agent to get its work done, but is not in and of itself an agent.
4. I am very reluctant to apply an authoritative name, because there are several valid ways to name them. However, I changed the scenarios to be numbered, and then added a paragraph discussing appropriate names for each one. See what you think.
5. Agreed. Added some sentences about discoverability and connection in the middle of "How Agents Talk".
6. Done. Good idea.
7. Done.
8. Done.
9. Done.

swcurran:
Awesome document with great writing.  It's technically helpful and correct, and the style and writing is extremely accessible.

I'm a little worried that for it's intended audience - a developer wanting to get started on developing SSI code - the Reference section raises way more question than it answers.  There are a ton of very subtle points raised in that section that could overwhelm someone new, and I don't think help a new developer.  Perhaps replace it with a link to another document with a note - "yes, there be dragons, and once you know  enough, this will help you".

I would consider adding some more to the general patterns section, focusing specifically on the two scenarios - Alice's agents and Faber's issuing agent.

- how agents connect - person to person, person to enterprise
- pairwise DIDs and where the ledger comes into play - always a question from new devs
- how Alice's two agents work together
- a bit more on how a Faber issuer might behave - this is partially covered but nice to have separated
- message families/protocols

Adding those in the same style would be awesome.  Great work!

mwherman2000:
@dhh1128, unpacking your reply to 3...
> You can't participate in the agent ecosystem without keys. It is a universal requirement. 

Agree ...for self-sovereign identities ...but this document about agents, characteristics of agents, and type of agents.

> You can hold a key without a wallet (e.g., in a config file). This is how static agents were defined. 

Agree

> You can't get so lightweight that you don't use keys. 

Agree ...for self-sovereign identities ...but this document about agents, characteristics of agents, and type of agents (repeated from above).

> An entity that doesn't hold keys can call an agent to get its work done, but is not in and of itself an agent.

What do you mean, specifically, by _entity_?

Lastly (related to last point), lightweight apps/clients/nodes typically have no local state (or virtually no local state).  In the context of the SSI Lightweight Agent, I'm assuming this to be true as well ...that the Lightweight Agent has no local state: no local wallet and not local configuration state (related specifically to the management of identities).  @dhh1128 I think we differ on this. ...here's the v0.5 of the chart...

#iDIDit: An Architecture-driven Taxonomy for SSI Agents v0.5

Reference: https://hyperonomy.com/2019/02/01/architecture-driven-taxonomy-for-ssi-agents/

dhh1128:
@swcurran Agreed. I will move the "here be dragons" stuff into a separate doc.

dhh1128:
@mwherman2000 The Indy community has chosen to define an agent using the 3 characteristics that I listed. This is not controversial; it has been settled for a long time. We know that "agent" has other definitions in other circles; in the intelligence community, they think about James Bond when they hear that word. In Archimate, there may be a different definition as well. But this is an SSI ecosystem, not a general-purpose enterprise software ecosystem. This is why an early paragraph of the doc says `When we use the term "agent" in the SSI community, we more properly mean "an agent of self-sovereign identity." This means something more specific than just a "user agent" or a "software agent."`

The reason why we chose our definition is that cryptographic keys are central to many constructs in the ecosystem. Keys are the unit of control and delegation for DIDs. They are also the unit of revocation for cred defs and agents. They are the unit of addressing and encryption when a message is packaged for one or more destinations. Keys govern credential issuance. Keys are the unit of signing. We want agents to align with these constructs, not span them in unhelpful ways.

Agents can't prove they hold delegated authority (my characteristic #1) without keys. So if you say that keys are not a defining characteristic of agents, you also eliminate delegation as a formally programmable/testable property. Furthermore, non-key-holding entitites can't participate in protocols either (except as an unknowable, invisible, unidentifiable entity fronted by something that *does* have keys), since all protocols depend on knowable participants, and knowable participants depend on encryption, which requires keys. So without keys, you can't have characteristic 3, either.

An app that doesn't hold keys cannot be held accountable for its own actions as a discrete participant in A2A interactions. An unaccountable piece of software is one that cannot be certified to comply with a trust framework and cannot be audited (e.g., to show that it generated a proof or sent a message). From A2A's perspective, it *isn't* a participant in interactions at the level of the protocol.

There is NO problem with building software that doesn't hold keys. Such software can add tons of business value and can act on the digital landscape. I'm sure there will be lots of stuff that works that way. But that stuff won't be an agent. We can't call it that, without completely reinventing a mental model that has huge momentum.

The thing you are calling a lightweight agent on a mobile device, I would call a tool or app or UI, but only half of an agent in the sense that our ecosystem expects. The agent is the combination of the front end on mobile, and the back end on cloud; together it exposes one discrete interaction surface to other participants, and its boundaries do not have to match hardware boundaries.

Static agents, which you placed in the "lightweight" bucket, DO hold keys. They just hold a statically configured set of them, and they don't negotiate new relationships. It is the staticness that defines them, not lack of keys.

mwherman2000:
@dhh1128 Apologies for triggering a lengthy reply.  Based on your reply, what do you think of this?

image

mwherman2000:
Idea/Suggestion for re-crafting the Complexities section: After working through various versions of the above model, I'm wondering if it makes sense to craft the Complexity section using a Capabilities based model.  For example, the list of capabilities that an Agent can assume might include:
- UsesLocalWallet or HasLocalWallet
- UsesLocalConfig
- UsesLocalLedger
- IsRelay
- IsMediator

- EdgeLocated
- CloudLocated

etc. etc. etc.

dhh1128:
@mwherman2000 Yes, I think that new diagram is great. The only thing I would change is that I think client apps could have local config. Not a local wallet, but local config.

mwherman2000:
> 1. I think client apps could have local config. Not a local wallet, but local config.

I was envisioning Client Apps to be the front-end thingy in your first reply @dhh1128 : "The agent is the combination of the front end on mobile, and the back end on cloud; together it exposes one discrete interaction surface to other participants, and its boundaries do not have to match hardware boundaries."

image

PR #87, opened 2019-02-06
FCP HIPE: Ursa Migration

mikelodder7:


mikelodder7:
> @mikelodder7 I think this looks fine. My one question is: what happens as Ursa grows richer, in terms of the way it might impose an indirect burden on Indy as far as dependencies are concerned? For example, if Ursa picks up support for Goldilocks because some other HL project wants it, is there a way that Indy can avoid incurring an indirect dependency on a Goldilocks library? We want a foundation with lots of features, but we don't want it to cause all consuming layers to incur unnecessary dependencies.

Just like any library, we plan to version it. Indy can target a specific version and when they wish to move forward to a newer version they can update any APIs that might have changed.

To address the dependency issue, Ursa plans to have features that are plug and play. If Indy does not want an indirect dependency on say Goldilocks, they can just pick the features they want and not select the rest. This is a nice feature of Rust. Dave Huseby and I are writing an RFC for how this should work but I believe it will work like this. Indy's Cargo.toml file would indicate ursa = { version = "..." features=["BBS+", "Ed25519"]. This means only include those listed features and everything else will be removed.

swcurran:
Agree that this should be moved forward.

text/ursa/README.md
```diff
@@ -0,0 +1,77 @@
...
+Here is a short list of some required code changes
+
+
+|  | Indy-Crypto | Ursa |
+| - | ----------- | ---- |
+| Function Prefix | indy_crypto_ or indy_ | ursa_ |
+| Error Names | IndyCryptoError | UrsaCryptoError |
+
+
+Ursa is now available as a [crate](https://crates.io/crates/ursa). This means that
```
jovfer: Has it been published manually or by some CD pipeline?

jovfer:
@mikelodder7 @dhh1128 Should this HIPE cover details about artifacts distribution and movement CI/CD pipelines from Evernym/Sovrin infrastructure to somewhere else?
Missed CI/CD in general is a blocker to switch other Indy projects to URSA (in my mind)

text/ursa/README.md
mikelodder7: Right now its done manually. We are working with Hyperledger to get a CI/CD pipeline going.

mikelodder7:
> @mikelodder7 @dhh1128 Should this HIPE cover details about artifacts distribution and/or movement CI/CD pipelines from Evernym/Sovrin infrastructure to somewhere else?
> Missed CI/CD in general is a blocker to switch other Indy projects to URSA (in my mind)

I added a section about this.

PR #88, opened 2019-02-07
proposed HIPE: QuestionAnswer protocol

glowkey:
Signed-off-by: Douglas Wightman <douglas.wightman@evernym.com>


committed_answer_protocol.md
```diff
@@ -0,0 +1,123 @@
...
+valid_response. The valid_response.text field can also support localization as the nonce is signed and not the text 
+itself.
+
+The response message is then sent using the @sig message decorator:
+
+```JSON
+{
+  "@type": "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/committedanswer/1.0/answer",
+  "@thread": { "thid": "518be002-de8e-456e-b3d5-8fe472477a86", "seqnum": 0 },
+  "response.@sig": {
```
TelegramSam: it would be helpful to show the unsigned field form of the message first, and then the signed version after.

committed_answer_protocol.md
```diff
@@ -0,0 +1,123 @@
...
+
+### Optional Elements
+
+The "question_detail" field is optional. It can be used to give "fine print"-like context around the question and all of
+its valid responses. While this could always be displayed, some UIs may choose to only make it available on-demand, in a
+"More info..." kind of way. 
+
+@timing.expires_time is optional
+
+**"response_signature" field is optional.**
```
TelegramSam: the signature is optional, but not the field, right?

It may be worth including a `signed_response_requested` in the request.

committed_answer_protocol.md
```diff
@@ -0,0 +1,123 @@
...
+```JSON
+{
+  "@type": "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/committedanswer/1.0/question",
+  "@id": "518be002-de8e-456e-b3d5-8fe472477a86",
+  "question_text": "Alice, are you on the phone with Bob from Faber Bank right now?",
+  "question_detail": "This is optional fine-print giving context to the question and its various answers.",
+  "valid_responses" : [
+    {"text": "Yes, it's me","nonce": "<unique_identifier_a+2018-12-13T17:00:00+0000"},
+    {"text": "No, that's not me!","nonce": "<unique_identifier_b+2018-12-13T17:00:00+0000"}],
+  "@timing": {
```
TelegramSam: `@timing` and `@thread` need to be updated to `~timing` and `~thread`.

committed_answer_protocol.md
```diff
@@ -0,0 +1,123 @@
...
+The protocol begins when the questioner sends a `committedanswer` message to the responder:
+
+```JSON
+{
+  "@type": "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/committedanswer/1.0/question",
+  "@id": "518be002-de8e-456e-b3d5-8fe472477a86",
+  "question_text": "Alice, are you on the phone with Bob from Faber Bank right now?",
+  "question_detail": "This is optional fine-print giving context to the question and its various answers.",
+  "valid_responses" : [
+    {"text": "Yes, it's me","nonce": "<unique_identifier_a+2018-12-13T17:00:00+0000"},
```
TelegramSam: I wonder if `response_code` would be a clearer name for the function of `nonce`.

glowkey: I agree. I will update this.

committed_answer_protocol.md
glowkey: This actually represented an old way of thinking so I'll update it. The response_signature is not optional and if it's not included the response is invalid. I'll update this to be an explainer on invalid responses.

swcurran:
Could we get some more business context around this message?  Notably - why would I want to use this protocol, and once I have the signed answer, how would I use it in a audit situation?  On that latter question - that would be from both a business context (an example of why/how an audit might come about) and from a technical context - how would I use it in audit scenario with another party?

tplooker:
@glowkey if we were to say have the responder/answer sign not only the response but also the original question, could this bare some use for anywise relationships. I.e I ask a party a question and they respond with a commited answer that anyone else in the anywise relationship could trust? Plus this might also yield benefit for certain parties in an audit, i.e being able to link the response to the question in an irrefutable way?

glowkey:
@tplooker, I think that's a valid use-case but outside the scope of this particular protocol. I'm attempting to keep it as locked to a single 2-party "question/answer" non-repudiable use case as possible. 

I originally thought about signing the actual answer text (rather than its corresponding nonce) but felt like allowing for internationalization was more important.

glowkey:
@swcurran I will add some more detail around business context.

committed_answer_protocol.md
```diff
@@ -0,0 +1,129 @@
...
+valid_response. The valid_response.text field can also support localization as the response_code is signed and not the text
+itself.
+
+The response message is then sent using the ~sig message decorator:
+
+```JSON
+{
+  "@type": "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/committedanswer/1.0/answer",
+  "~thread": { "thid": "518be002-de8e-456e-b3d5-8fe472477a86", "seqnum": 0 },
+  "response.~sig": {
```
TelegramSam: This should be `response~sig`. The extra period (.) is improper.

TelegramSam:
I LOVE this HIPE, with one exception: I think there is value in this protocol not requiring a signed response. The same protocol can then be used for non-critical things like an office lunch poll. I realize that the name of the protocol would need to be changed, but It seems ridiculous to have two parallel protocols, with one requiring a signature and one not.

If the initial message indicates that a signature is required, then the response can be signed. Otherwise, it isn't. Thoughts?

glowkey:
Whether or not Faber requests a signature, Alice can choose for herself to sign the answer or not. It's then up to Faber to decide whether Alice's response is what it wants. If it's not signed then Faber can ask again or accept it unsigned. Thoughts?

TelegramSam:
Under the current design, there isn't a way to ask for a repudiable answer, only a non-repudiable one. This means the expectation (or request) for a signature is always present. If we make the signature an optional element of this protocol, it would serve both cases well. I belive this would require a protocol rename, but would be worth it.

glowkey:
I'm convinced and have renamed this to QuestionAnswer protocol and made signing optional. @TelegramSam I'm interested in your feedback.

dhh1128:
@glowkey Would you consider moving this over to the aries-rfcs repo? This sort of protocol now belongs there. Moving is fairly easy; see https://docs.google.com/document/d/1BKLR6lmjuwmHrYldNELijddJYi43IJByBXSMQgIHSYM/edit.

PR #94, opened 2019-02-26
Propose HIPE: Transports

TelegramSam:


text/transports/README.md
```diff
@@ -0,0 +1,94 @@
...
+- Name: (fill me in with a unique ident in kebab case, such as "my-awesome-feature")
+- Author: (your name and/or email)
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- PR: (leave this empty)
+- Jira Issue: (leave this empty)
+
+## Summary
+[summary]: #summary
+
+One paragraph explanation of the feature.
```
swcurran: @TelegramSam - I think it is a little too early in the development of this HIPE to consider accepting. Perhaps you could add some more detail?  :-)

TelegramSam: lol. Fixed. forgot to hit save before my commit. 🤦‍♂️

text/transports/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+### HTTP(S)
+
+HTTP(S) is considered the main transport for Agent Messaging.
+
+- Messages are transported via HTTP POST.
+
+- The MIME Type for the POST request is `application/ssi-agent-wire`.
+
+- A received message should be responded to with a 202 Accepted status code. This indicates that the request was received, but not necessarily processed. Accepting a 200 OK status code is allowed.
```
tmarkovski: Any reason why we can't have all 2XX status codes allowed here? As a transport definition, I feel it should allow for broader use of all *Success* status codes, including 201 Created, 204 No Content, etc.

kdenhartog:
What's the extent of error handling that we want to cover with the transport layer. For example, would a misformatted wire message return a 4XX status code? I'd think that it wouldn't make sense to use the status code to push this error back.  On the other hand, what about a message that is posted to the wrong endpoint? I would think that it would be acceptable to return a 4XX status code.

Also, given that we support IoT devices I think it's only fair to support the 418 status code if we're interacting with a teapot.

text/transports/README.md
TelegramSam: I think if we are to support a broader range of 2xx response codes, we should define what they indicate.

TelegramSam:
> What's the extent of error handling that we want to cover with the transport layer. For example, would a misformatted wire message return a 4XX status code? I'd think that it wouldn't make sense to use the status code to push this error back. On the other hand, what about a message that is posted to the wrong endpoint? I would think that it would be acceptable to return a 4XX status code.

I do think we need to be careful here to not step into full message processing. Errors related to message _receipt_ might be acceptable though.

> Also, given that we support IoT devices I think it's only fair to support the 418 status code if we're interacting with a teapot.

Definitely need to work out the particulars of this one. :)

text/transports/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Reference
+
+[Reference]: #reference
+
+Standardized transport methods are detailed here. HTTP(S) is considered the main transport.
+
+### HTTP(S)
+
+HTTP(S) is considered the main transport for Agent Messaging.
```
dhh1128: I strongly disagree with this statement. However, I could be comfortable with an alternative statement that "HTTP(s) is the first transport for DID Communication that has received heavy attention."

text/transports/README.md
```diff
@@ -0,0 +1,74 @@
...
+## Motivation
+[motivation]: #motivation
+
+Agent Messaging is designed to be transport independent, including message encryption and agent message format. Each transport does have unique features, and we need to standardize how the transport features are (or are not) applied. 
+
+## Reference
+
+[Reference]: #reference
+
+Standardized transport methods are detailed here. HTTP(S) is considered the main transport.
```
dhh1128: I don't agree that HTTP(s) is the main transport. The reason the community has focused there is because everybody is writing enterprise agents. For enterprises, it makes a lot of sense. It does NOT make a lot of sense for consumer edge agents, where the default transport is likely to be push notifications or sockets or Bluetooth.

This sentence doesn't really help in any way; I think we should just strike the sentence.

text/transports/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+- Messages are transported via HTTP POST.
+
+- The MIME Type for the POST request is `application/ssi-agent-wire`.
+
+- A received message should be responded to with a 202 Accepted status code. This indicates that the request was received, but not necessarily processed. Accepting a 200 OK status code is allowed.
+
+- POST requests are considered transmit only by default. No agent messages will be returned in the response. This behavior may be modified with additional signaling.
+
+- Using HTTPS with TLS 1.2 or greater with a forward secret cipher will provide Perfect Forward Secrecy (PFS) on the transmission leg.
```
dhh1128: This is true. However, the statement needs to be reconciled against the merged HIPE that says we will achieve perfect forward secrecy a different way.

dhh1128:
Aside from the comments I left, my other point of feedback is just that I think we need discussion of a few more issues and transports before we merge this. For example:

1. How are different transports tested?
2. How mediators and relays relate to transports (need hyperlinks to that HIPE)
3. How message trust contexts might be affected by transports (or not).
4. How to standardize handling of a new transport. (By raising a PR against this doc? If so, what questions would the PR have to answer?)
5. Which transports have known implementations.

text/transports/README.md
TelegramSam: good alternate. made that change.

text/transports/README.md
TelegramSam: striken

TelegramSam:
@dhh1128 I think I've addressed the points you mentioned. Any additional feedback is very welcome.

dhh1128:
@TelegramSam I believe this PR is now superseded by Aries RFC 0025, so this PR can be closed. Correct?

TelegramSam:
superseded by https://github.com/hyperledger/aries-rfcs/tree/master/features/0025-didcomm-transports

PR #95, opened 2019-02-26
DRAFT HIPE: Ephemeral challenge protocol

tplooker:
Initial draft HIPE, one of a series up and coming.


text/ephemeral-challenge-protocol/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+#### Challenge message
+
+The protocol can begin when the challenger sends a `challenge` message to the other party which looks like this:
+
+```
+{
+    "@type" : "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/ephemeral_challenge/1.0/challenge",
+    "@id" : "624f736c-bbb1-4236-945f-8a9feff012dd",
+    "recipientKeys": ["8HH5gYEeNc3z7PYXmd54d4x6qAfCNrqQqEB3nS7Zfu7K"],
```
swcurran: Why not just use the diddoc format for this versus something different?  For example, in this, we don't specify the type of the keys, but presumably that is important (necessary?).  Since we know a properly formatted, minimal diddoc will always be correct, why not just use it whenever we need to convey a "public-key-plus-service-endpoint" structure?

text/ephemeral-challenge-protocol/README.md
```diff
@@ -0,0 +1,159 @@
...
+    "routingKeys": ["8HH5gYEeNc3z7PYXmd54d4x6qAfCNrqQqEB3nS7Zfu7K"]
```
swcurran: Could a public DID be used that is resolved by the Challenge Accepter?

text/ephemeral-challenge-protocol/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Use Case Description
+
+### Alice purchasing an age restricted product online (example using QR codes as out of band process)
+
+Alice has reached checkout with an online retailer `wine-to-go`, where she wishes to complete purchase for alcohol which is age restricted. Alice has already got a mobile wallet and has obtained a credential (such as a driver licence) from a trusted issuer. Alice has no prior connection with `wine-to-go`.
+
+1. On the checkout page of `wine-to-go`, Alice clicks the `Verify age with mobile wallet` button.
+2. A QR code appears with instructions for Alice to scan it using her mobile wallet.
+3. Using her mobile wallet Alice scans the code and a prompt in the app appears saying `Verification required` with the information indicating that `wine-to-go` is requesting proof of her age.
```
swcurran: Based on the data passed below - how does Alice see on the screen it is "wine-to-go" asking?  I don't think that is in the proof request, is it?

text/ephemeral-challenge-protocol/README.md
tplooker: Essentially I have just copied the same format as the invite here, could be convinced that a did doc representation could be added.

text/ephemeral-challenge-protocol/README.md
tplooker: Potentially, if we are happy with a split rule i.e `routingKeys` is an array of base58 encoded keys or did key references?

text/ephemeral-challenge-protocol/README.md
tplooker: This use case was really trying to frame this HIPE and isnt enough alone to describe the full picture, I was merely trying to give it a bit of context, once I have added the other HIPES in conjunction it should make more sense :)

TelegramSam:
I'd like to drill down into why a connection may not be desired. Not contesting it yet, but I think we can learn a bit by the discussion.

swcurran:
I think it is necessary based the number of connections that a person has - the volume of connections.  I have about 800 entries in my lastpass repo, so that is at least how many legit connections that a person might have, never mind trips to the liquor store, and pubs.

tmarkovski:
A connection makes this challenge flow obsolete, as it is authentication in itself, with it's own challenge/response flow (connection protocol). I think the intention with this hipe's flow is to reduce the friction of connection establishment and allow transient authentication without establishing permanent connection. I think there are a lot of applications for ephemeral challenge, especially over offline networks (NFC, Bluetooth, etc).
@swcurran makes a good point, we don't want a connection entry for every authentication endpoint unless it's necessary for credential exchange, not just presentation.

dhh1128:
After reading this HIPE draft, I am feeling like I need a better definition of the problem this protocol is solving.

What do we mean by "ephemeral" here? "Connectionless"? If so, we should use "connectionless". Ephemeral means transitory or temporary, not connectionless.

I also don't know what we mean by "challenge." Are we imagining many different types of challenge can be plugged in? If so, how does that work? The only example I am seeing is a challenge for a proof request. So I am wondering if this is really the connectionless proof request protocol, and not the ephemeral challenge protocol.

tplooker:
@dhh1128 Sorry I totally missed the comment on this. 

I called the protocol ephemeral challenge as it is an "instantaneous" challenge protocol (i.e no prior connection needed to host the protocol and no commitment passed responding to the challenge).

My thinking about why I called it a "challenge" protocol rather than a "proof request" protocol is that yes, I think in the very near future, even though I have not articulated this yet in the HIPE (hence it is only a draft) that there will be other forms of challenges. The most notable could be a pairwise challenge for example the challenger generates a challenge signed with their public did, a challenge responder receives and identifies the public did is linked to a peerwise connection they have in their wallet and responds with a challenge response that identifies them to the challenger. Although in the second challenge scenario I have outlined, a connection is required. This is not an artifact of the protocol it self rather the specific type of challenge creating that requirement.

Its important to note that this protocol has a large overlap with did-auth and I am still assessing its overlap.

In general what I am hoping to accomplish with this protocol is a general purpose challenge request-response protocol.

tplooker:
@dhh1128 and any other interested parties, please take a look at the updated HIPE this is still a draft mainly designed to promote discussion around the concept. I have added an early example as to how a `proof of pairwise connection` could work.

text/ephemeral-challenge-protocol/README.md
```diff
@@ -0,0 +1,237 @@
...
+
+"Ephemeral" - The protocol is largely about instantaneous interactions. Even though a challenge issued in this protocol could have a reliance on prior state such as the existence of a connection or a credential in the challenge responders wallet.
+
+"Challenge" - The protocol is about arbitrarily challenging a party, whether that be to prove possession of a verifiable credential, or a key relating to a persisted connection. 
+
+## Use Case Description
+
+### Alice logging into her bank Faber (example using QR codes as out of band process)
+
+Alice has become a customer of Faber, in the process of becoming a customer, Faber connected with her SSI wallet and created a persisted pairwise connection with her. During the setup of this connection, Faber also disclosed to Alice its public DID used to publicly identify the organization (`did:sov:abcdefg12345`) and Alice associated this DID to the pairwise connection she has with Faber.
```
dhh1128: I don't believe this is valid behavior. did:sov:abcdefg12345 is an anywise DID, not a pairwise DID. Alice should use a standing connection invitation, published on Faber's website, to create a pairwise relationship with Faber. See https://docs.google.com/presentation/d/1UnC_nfOUK40WS5TD_EhyDuFe5cStX-u0Z7wjoae_PqQ/edit#slide=id.g46da93b149_0_145

text/ephemeral-challenge-protocol/README.md
```diff
@@ -0,0 +1,237 @@
...
+1. Alice visits `www.faber-bank.com`.
+2. Alice navigates to the login page of the Faber bank website and clicks the `Login with SSI` button.
+3. A QR code appears with instructions for Alice to scan it using her mobile wallet.
+4. Using her mobile wallet Alice scans the code and a prompt in the app appears saying `Verification required` with the information indicating that Faber Bank is requesting proof of connection.
+5. Alice accepts this request on her mobile app and notices that the Faber bank website has updated and logged her into her account.
```
dhh1128: Because I don't buy your starting conditions, I don't believe in the use case at all. Login should happen when connections exist. It shouldn't be fit into a connectionless flow.

text/ephemeral-challenge-protocol/README.md
```diff
@@ -0,0 +1,237 @@
...
+
+### Alice purchasing an age restricted product online (example using QR codes as out of band process)
```
dhh1128: I am also pretty dubious about this use case. It is fine as far as it goes--but as soon as Alice has to check out, won't she establish a connection in order to provide payment and shipping info? Or do you imagine that those will be handled in connectionless fashion as well?

I think there *are* good use cases for this protocol; I just don't think these are very compelling. Better ones are likely to involve physical and/or temporal proximity, because those ingredients make the lower level of trust more reasonable.

text/ephemeral-challenge-protocol/README.md
```diff
@@ -0,0 +1,237 @@
...
+
+#### Null
+
+No challenge exists
+
+#### Challenged
+
+A challenge has been generated by the challenger and shared with a challenge responder
+
+#### Challenge Accepted
```
dhh1128: Wouldn't we need a "Challenge Rejected" response, too? "Sorry, I don't feel like proving to you that I'm over 21." Or is that done with a problem-report?

text/ephemeral-challenge-protocol/README.md
```diff
@@ -0,0 +1,237 @@
...
+
+`recipientKeys` an array of base58 encoded keys to which the challenge response should be encrypted to using `anoncrypt`
+
+`routingKeys` an array of base58 encoded keys to which the message should be wrapped for
+
+`serviceEndpoint` the uri to which the resulting packed challenge response should be sent to
+
+`challenge` the challenge, see the challenge types section for more information.
+
+Note : It is important to note that how the challenge message is communicated to the challenger is out of band
```
dhh1128: If the challenge is communicated out of band, then it has no security guarantees that derive from A2A. Thus, a MITM could substitute their own challenge. This could be okay, when physical and/or temporal proximity provide extra protections--but it is unlikely to be okay with arbitrary remoteness.

text/ephemeral-challenge-protocol/README.md
```diff
@@ -0,0 +1,237 @@
...
+
+#### Indy Proof Request
+
+An indy based proof request is when the challenger is asking for proof of certain credentials 
+
+The challenge field in the challenge message takes the following form.
+
+```
+challenge : {
+    "type" : "indy-proof"
```
dhh1128: Use embedded attachments? Instead of "type", "mime-type"? Are we reinventing the wheel?

text/ephemeral-challenge-protocol/README.md
```diff
@@ -0,0 +1,237 @@
...
+```
+"response" : {
+    "type" : "indy-proof",
+    contents: {
+        //libindy proof
+    }
+}
+```
+
+#### Pairwise Connection Proof
```
dhh1128: A simple trust ping-response accomplishes exactly the same thing. All we would have to do is deliver it to an arbitrary endpoint, signed as you specify. In other words, it is not the message format that's valuable here; it's the signing and the arbitrary delivery. So why invent a new message type?

text/ephemeral-challenge-protocol/README.md
tplooker: What I am trying to communicate here is that Alice has two dids for Faber bank, the pairwise did that Faber shared with her when they connected and fabers public did or anywise did that makes Faber identifiable in a public situation. 

All communication Alice->Faber would use the pairwise did for encrypting messages but any public challenges Faber issues, say as a challenge in browser would be signed by Fabers public did.

text/ephemeral-challenge-protocol/README.md
tplooker: I agree on the back of our conversation the other day, making the challenge protocol a more arbitrary mode of operation for agents makes sense. My only reason for updating the HIPE to include this was to convey that this protocol was not too be limited to proof presentation and that other challenged based use cases exist.

dhh1128:
@tplooker Do you think we can close this PR and raise a more evolved version of the ideas in a new Aries RFC PR?

tplooker:
Yes I will raise a new one in aries

PR #96, opened 2019-02-27
DRAFT HIPE: Oauth ephemeral challenge protocol integration

tplooker:
This is a very early draft of how the OAuth integration is intended to work with the ephemeral challenge protocol, there is much still to discuss, such as the `ephemeral challenge message` being signed inside a JWS structure.


text/ephemeral-challenge-protocol/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+The protocol can begin when the challenger sends a `challenge` message to the other party which looks like this:
+
+```
+{
+    "@type" : "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/ephemeral_challenge/1.0/challenge",
+    "@id" : "624f736c-bbb1-4236-945f-8a9feff012dd",
+    "recipientKeys": ["8HH5gYEeNc3z7PYXmd54d4x6qAfCNrqQqEB3nS7Zfu7K"],
+    "routingKeys": ["8HH5gYEeNc3z7PYXmd54d4x6qAfCNrqQqEB3nS7Zfu7K"]
+    "serviceEndpoint": "https://example.com/endpoint",
```
TelegramSam: It strikes me that the routing information (or a public DID) could be attached to nearly any message. Perhaps a decorator is the right way to embed this into a message.

tplooker: @TelegramSam Interesting you say this as I was thinking along similar lines this morning, I think what ever format is agreed upon here should be the same as what is present in the invite?

dhh1128:
@tplooker Do you think we can close this PR and raise a new, more evolved version of the thinking in an Aries RFC PR?

tplooker:
Yes I will raise a new one in aries

PR #99, opened 2019-03-05
tweak a2a-->DIDComm

dhh1128:


swcurran:
This question should be finalized on the next Indy Maintainers call:

> Is "DIDComm" the accepted term going forward?

If so, this should be merged immediately.

PR #100, opened 2019-03-08
FCP HIPE: repudiation explainer

dhh1128:


kdenhartog:
Thanks for adding this documentation. An important premise that I realized when discussing this with members in the community is that repudiable integrity of data as the default maps very well to the real world. 

For example, when I'm in public I don't have a microphone following me around recording my every conversation and then using that to prove I said something. Only in very specific instances do we put things "on the record" so to say. In instances like public interviews, signing documents, and legal testimonies it's useful. However, we don't have an expectation that every conversation is recorded (ignoring the fact that I carry a cell phone with me), so it make sense that our technologies extend that paradigm.

PR #105, opened 2019-03-19
Cross-Domain Messaging: Changed format of "msg" field.

Artemkaaas:
Motivation:
Libindy `pack` function in the current implementation builds JSON object and then does a conversion to bytes.
- Setting bytes as is to A2A `msg` field is super inefficient because of the following message JSON serialization.
- Applying additional base64 wrapping increase message size also

Solution: There the next possible options:
- converting the object back from bytes (as I suggested in PR) before setting as a value
- converting bytes to string some simple way
- changing of `pack` function to return JSON string instead of conversion to bytes.

Signed-off-by: artem.ivanov <artem.ivanov@dsr-company.com>


dhh1128:
@Artemkaaas , is your change here compatible with the change that @lovesh is recommending in this PR against indy-sdk? https://github.com/hyperledger/indy-sdk/pull/1537

Artemkaaas:
@dhh1128, it's a different level. @lovesh suggested changes inside libindy `pack` function (how JSON object is built).

jovfer:
@dhh1128 this PR is just for clarification current approach and probably for minimal instant changes.

BTW I expect large update from @lovesh and large discussion around topic "is it required and how to avoid re-encryption and re-base64 data while multiply rounds of pack and fwd"

TelegramSam:
Can we update this to clarify that the msg is not a string but the json object? IE we want this: 
`"msg": { "protected": "...", ... }`
and not this:
`"msg": "{ \"protected\": \"....\", .... }"`

dhh1128:
@TelegramSam I am thinking this is ready to merge. Do you agree?

PR #110, opened 2019-03-29
propose HIPE: Introductions

dhh1128:


text/introductions/README.md
```diff
@@ -0,0 +1,323 @@
...
+the scope of the introduction.
+
+This suggests an important insight about the relationship between the
+introduction protocol and the [connection protocol](
+https://github.com/hyperledger/indy-hipe/blob/master/text/0031-connection-protocol/README.md):
+*they overlap*. The invitation to form a relationship, which
+begins the connection protocol, is also the final step in an
+introduction.
+
+Said differently, *the goal of the introduction protocol is to start the
```
swcurran: Clarification is needed to define what the intended target relationship between the participants for a specific instance of an introduction.  There are two possibilities that I see:

* Create a new relationship between Bob and Carol
* Create a new relationship between Bob, Carol and Alice

Arguably, a third relationship could be to add Bob to Carol and Alice's existing relationship, but I would argue that might be the connection protocol or (I think more likely) a new protocol that we've not defined.  I would suggest not including it in this protocol.

Suggestion is to make it clear which target relationship is intended as you go through this. I think throughout you are talking about the first case - a new Bob and Carol relationship.

text/introductions/README.md
```diff
@@ -0,0 +1,323 @@
...
+
+Bob and Carol may actually know each other already, without Alice realizing
+it. The introduction may be rejected. It may create a new pairwise
+relationship between Bob and Carol that is entirely invisible to Alice.
+Or it may create an n-wise relationship in which Alice, Bob, and Carol know
+one another by the same identifiers.
+
+Some specific examples follow.
+
+#### One introducee can't do DIDComm
```
swcurran: I don't get this use case in the introduction protocol context.  If Bob doesn't have a DIDComm capability then Alice doesn't have an existing DIDComm-based relationship to use for the introduction protocol.  That would be addressed by Alice using the connection protocol to connect with Bob so he does have DIDComm capabilities.

text/introductions/README.md
```diff
@@ -0,0 +1,323 @@
...
+In the tutorial narrative, only a simple proposal was presented. A
+fancier version might be:
+
+[![fancy proposal example](fancy-proposal.png)](fancy-proposal.json)
+
+This adds a number of fields to the introducee descriptor. Each is optional
+and may be appropriate in certain circumstances. Most should be self-explanatory,
+but the `proposed` field deserves special comment. This tells whether the
+described introducee has received a proposal of their own, or will be
+introduced without that step.
```
swcurran: I don't understand this example (why are we scheduling an MRI in an introduction?), nor the `proposed` field. A further explanation would be helpful.

text/introductions/README.md
```diff
@@ -0,0 +1,323 @@
...
+
+This example also adds the `nwise` field to the proposal. When `nwise` is
+present and its value is `true`, the proposal is to establish an nwise
+relationship in which the introducer participates, as opposed to a pairwise
+relationship in which only the introducees participate.
```
swcurran: I don't understand the implications of this.  What changes when `nwise` is `true`?  That would seem to require a lot to the protocol.

text/introductions/README.md
```diff
@@ -0,0 +1,338 @@
...
+##### Transferring Trust
+
+[TODO: talk about how humans do introductions instead of just introducing
+themselves to strangers because it raises trust. Example of Delta Airlines
+introducing you to Heathrow Airport; you trust that you're really talking
+to Heathrow based on Delta's asertion.]
+
+### Roles
+
+There are three [TODO:do we want to support introducing more than 2 at a time?]
```
tplooker: In AgentFramework we have the concept of a multi-party connection invitation, i.e I might be doing a presentation where I put a QR code at the end of the presentation for any attendees to connect with me after. Perhaps this should be a consideration in this protocol for multi-party introductions?

dhh1128:
This has been superseded by Aries RFC 0028.

PR #112, opened 2019-04-04
Improved handling of nested packed messages

lovesh:
Signed-off-by: lovesh harchandani <lovesh.bond@gmail.com>


kdenhartog:
I'm hesitant to the ideas being proposed here. I understand the concern of multi-wrapping and the inefficiency of it, and agree we will need a solution to scale this. However, with that said I want to hold off on this until we've gotten others to accept the current "profile" of the JWE. 

Based on rough numbers [here](https://joshzeigler.com/technology/web-development/how-big-is-too-big-for-json), we can support a system of about 3-5 wraps without encountering large issues in a browser in terms of DOM load time.Additionally, onion wrapping is not the only practical way to keep the size down. We also discussed an idea of JIT encrypting.

With regards to the actual API additions, I would rather include this into the pack_message/unpack_message API rather than create additional APIs. The primary reason for this is to reduce the complexity for a developer who is programming an agent that is receiving a message. Right now unpack handles the complexity of this, which removed the poor developer experience that Anoncrypt and Authcrypt brought.

Given these two concerns, I would like to hold off until we can find broader alignment on wire level messages. My expectation is that come the end of April (IIW) we can have a better understanding if we should standardize on this approach within Indy. With that said, we should still look at what it would take to add the support into the pack and unpack message rather than adding additional APIs.

lovesh:
> We also discussed an idea of JIT encrypting.

Does it mean Just-in-time? Can you point to a link describing this or describe here a bit about this?

> I would rather include this into the pack_message/unpack_message API

The objective was to not change pack_message/unpack_message API as they need to be JWE compliant. But i agree that if pack/unpack had this, it would be a better developer experience.

> With that said, we should still look at what it would take to add the support into the pack and unpack message rather than adding additional APIs.

It does not take much coding work to move this functionality to pack/unpack but it might take a lot of convincing to have an exception/addition in the JWE spec

kdenhartog:
> Does it mean Just-in-time? Can you point to a link describing this or describe here a bit about this?

Yes, it's Just in time. An example of how this might work would be (using the numbers from the diagram in this HIPE) 

ENC1(plain, 1, 4) -> DEC2(ENC1, 1, 2) -> DEC2(ENC2, 2, 1) -> ENC2A(ENC1, 2, 8) 
-> DEC2A(ENC2A, 8, 2) -> ENC2B(ENC1, 8, 9) -> DEC2B(ENC2B, 9, 8) -> ENC2C(ENC1, 9, 4) -> DEC2C(ENC2C, 4, 9) -> DECplain(ENC1, 4, 1) -> plaintext

Essentially what is happening is that the encrypted plaintext gets reencrypted multiple times. This still protects the sender's anonymity preventing intermediaries from knowing who's talking, while at the same time allowing the intermediaries to know who to send the message to next. Additionally, because the edge agent can sign messages with the receiving mediator's endpoint, signifying that when ever this key is seen it should be forwarded to the receiving mediator's endpoint, the sending intermediary won't know who is communicating.  

> The objective was to not change pack_message/unpack_message API as they need to be JWE compliant. But i agree that if pack/unpack had this, it would be a better developer experience.
> It does not take much coding work to move this functionality to pack/unpack but it might take a lot of convincing to have an exception/addition in the JWE spec

I'll address these two comments together. By changing the format of the wire level message format (adding the ciphertext section), we essentially are changing pack and unpack format. I say this because if we believe that wire message format is the "base" protocol that everyone operates from, then we need broad support. We can't have some people doing efficient wrapping, and others supporting inefficient wrapping and not create a complexity issue for ourselves, so I think if we decide this is the route to go, then we should seek the path of having an exception/addition/spec written that is broadly accepted by the broader community.

jovfer:
+1 for Kyle's feedback. Especially I don't like the idea to have a lot new API calls. I've requested changes on implementation PR https://github.com/hyperledger/indy-sdk/pull/1574 to avoid merging PoC implementation before finish initial discussion here

text/0028-wire-message-format/README.md
```diff
@@ -312,6 +312,172 @@ For a reference implementation, see https://github.com/hyperledger/indy-sdk/blob
...
+            {
+                "iv": "<some IV>",
+                "tag": "<some tag>",
+                "ciphertext": ENC( m ),
+            }
+        ]
+    }
+    ```
+    `collapse_ciphertext` is idempotent, meaning repeatedly applying this method on packed messages will lead to the same result. This is helpful in scenarios where `pack_msg` is applied in succession to a message without any other transformation in between like `pack_msg( collapse_ciphertext( pack_msg( msg ) ) )`
+    *An alternate that has been discussed is to omit `iv`, `tag` and `ciphertext` from the packed message and it is implicitly understood that they refer to the last element of `~cyphertexts`. The concern here is that it means we are more non-compliant with JWE spec.*
```
jovfer: Let me share some draft alternative with better JWE spec compliant rather current HIPE proposal:
=============
Initial message M0:
```json=
{
    "@type": "some msg type"
}
```
M0 packed -> P0:
```json=
{
    "protected": ...,
    "tag": ...,
    "iv": ...,
    "@ct": ct0
}
```
P0 prepared to forward -> F1:
```json=
{
    "@type": "fwd",
    "to": "toDID",
    "ct_meta": {
        "protected": ...,
        "tag": ...,
        "iv": ...
    },
    "@ct": ct0
}
```

F1 packed again -> P1:
```json=
{
    "protected": ...,
    "tag": ...,
    "iv": ...,
    "@cts": [ctF1, ct0]
}
```
while this packing `ct0` just copied to `@cts` and ignored while encryption
==========
and update pack logic to be able to deal with `@cts` array

lovesh: But we already have pack as JWE compliant. The API to avoid re-encryption can be used by someone who wants the efficiency but does not need *immediately* care about being JWE compliant. Updating pack with these changes means we have to update codebases already using pack and spend more time convincing about this new format. 
I would suggest we accept this extension (including PR) for the meantime so that devs needing it can use it. We can deprecate these new methods later once we have convinced other folks to accept this functionality as part of pack.

dhh1128:
FWIW, I like Lovesh's approach, and I believe merging it quickly is better than merging it later, because it avoids us having to support two implementations.

kdenhartog:
@dhh1128 and I got a chance to discuss this further. We believe first and foremost we should discuss this on an agent call (preferably before IIW) in hopes of having our community come to a conclusion. If we decide we (the indy community) find it more useful to bring this in than we can move forward, otherwise we'll wait and discuss this with the broader SSI community.

From an implementation standpoint, we both thought that it would be better for pack to handle this with an additional API parameter and then on the unpack side, we would have the unpack message parse and handle both formats. This would allow us the broadest support (Single JWEs only, Multi wrapped JWEs, and single JWE with efficient multiwrapping) while also taking into consideration the developer perspectives in our opinions. 

What's your thoughts on this approach @lovesh @jovfer 

Also, @TelegramSam could we get some time to discuss this on the agent call before IIW?

dhh1128:
I'm just adding my comment to second what @kdenhartog said. I think the same functionality but bundled into existing APIs is a net improvement. And Kyle helped me see why it would make sense to socialize a little more before merging--not so much because the ideas aren't what we want, as because we need to get buy-in and that must be managed delicately.

lovesh:
Alright we can add a `collapse` parameter to `pack` that would do the job of applying `collapse_ciphertext` on the result of `pack` and unpack will use `expand_ciphertext` if it finds the ciphertexts array.
But any opposition to keeping `add_cts_to_msg` and `remove_cts_from_msg` since they allow stripping the ciphertexts array and putting it back on. We need them when creating messages like Forward otherwise any subsequent `pack`s have to inspect the message content?

dhh1128:
@lovesh I'm not sure why we need these extra methods. Couldn't pack() create `~ciphertexts` (if a new boolean param called `repack` is false) or add to existing `~ciphertexts` (if `repack` is true)? And couldn't unpack keep pulling items out of ~ciphertexts (if the value of `ciphertext` is "$0")?

lovesh:
@dhh1128 The `collapse` argument does what you are suggesting with `repack`. The 2 methods i am proposing to keep are used when the result of pack is used in another message like a Forward. Eg. Alice `pack`s message `m` with `collapse` as True as 
```
{
    "protected": "<receiver info>",
    "iv": "$0",
    "tag": "$0",
    "ciphertext": "$0",
    "~cyphertexts": [
        {
            "iv": "<some IV>",
            "tag": "<some tag>",
            "ciphertext": ENC( m ),
        }
    ]
}
```

Now this message needs to be put in a `Forward`. The resulting `Forward` is this
```
{
    "@type" : "....forward",
    "to"   : "did:sov:1234abcd#4",
    "msg": {
        "protected": "<recipient info>",
        "iv": "$0",
        "tag": "$0",
        "ciphertext": "$0",
        "~cyphertexts": [
           {
             "iv": "<some IV>",
             "tag": "<some tag>",
             "ciphertext": ENC( m ),
           }
        ]
    }, 
}
```

When this Forward is passed to `pack`, `pack` has to look inside `msg` element of Forward (hence it has to know the semantics of `Forward`) and ignore `~cyphertexts`, perform the encryption and then populate `~cyphertexts` depending on whether `collapse` was true or not.

What this HIPE proposes is to keep `~cyphertexts` as a meta element and not process it during any other operation than collapsing the ciphertext. Hence the `Forward` is created from the packed message after stripping `~cyphertexts` using `remove_cts_from_msg` as 
```
{
    "@type" : "....forward",
    "to"   : "did:sov:1234abcd#4",
    "msg": {
        "protected": "<recipient info>",
        "iv": "$0",
        "tag": "$0",
        "ciphertext": "$0"
    }
}
```

The ciphertext is added back with `add_cts_to_msg` resulting in 
```
{
    "@type" : "....forward",
    "to"   : "did:sov:1234abcd#4",
    "msg": {
        "protected": "<recipient info>",
        "iv": "$0",
        "tag": "$0",
        "ciphertext": "$0"
    }, 
    "~cyphertexts": [
        {
            "iv": "<some IV>",
            "tag": "<some tag>",
            "ciphertext": ENC( m ),
        }
    ]
}
```

Now when this is passed to `pack`, it knows that the outermost `~cyphertexts` is not to be encrypted. `pack` would follow this rule for all messages

text/0028-wire-message-format/README.md
```diff
@@ -312,6 +312,172 @@ For a reference implementation, see https://github.com/hyperledger/indy-sdk/blob
...
+        "iv": "<some IV>",
+        "tag": "<some tag>",
+        "ciphertext": ENC( m )
+    }
+    ```
+    `collapse_ciphertext` will return 
+    ```
+    {
+        "protected": "<recipient info>",
+        "iv": "$0",
```
jovfer: I would like to share my thoughts with @dhh1128 and @kdenhartog . It has already discussed with Lovesh briefly, but I would like to have broadly discussion.

I don't see particular value to have reference like `"iv": "$0"`. My feeling it's better to either omit these fields or use something like https://github.com/hyperledger/indy-hipe/pull/112/files#r273393385
It makes the format more clear for understanding and may simplify implementation, especially in case of extending pack/unpack behavior (instead of making new calls)

dhh1128: This is true. However, it makes the format diverge from the format of JWEs.

jovfer: Both approaches diverge from JWEs, I think.

kdenhartog: This is correct, only the inefficient method of wrapping would be close to the JWE spec. I'll be able to start the discussions about this in DIF hopefully this week, and if not it will be at IIW next week. I'll provide updates when I've gotten a chance to bring this up.

kdenhartog:
I'm in favor of finding a way to optimize the pack method now, given that Hubs are looking to use a multi wrapping scheme. However, I haven't had time to discuss this idea further yet. I'd be in favor of closing this PR (since it makes sense as an Aries-RFC now) and opening an issue over there to discuss this further. I think there's good thinking here that will require some level of tweaking, but once we find a solution I'll PR it onto [Aries RFC #0019](https://github.com/hyperledger/aries-rfcs/pull/19)

kdenhartog:
lots of discussion about this work since it was originally proposed. I suggest that this PR be closed and the discussion is moved to Aries-rfc repo where most of the discussion has been occurring.

PR #113, opened 2019-04-08
Streamlined process

dhh1128:
A simpler process...


README.md
```diff
@@ -1,242 +1,120 @@
...
+  
+Make sure that all of your commits satisfy the [DCO requirements](
+https://github.com/probot/dco#how-it-works) of the repo.
+  
+### How to get a HIPE accepted
+
+After your HIPE PR is raised, the HIPE will receive feedback from the larger
+community, and the author should be prepared to revise it.
+
+When you believe that the HIPE is mature enough (feedback is resolved,
```
swcurran: It would be nice to have a comparison of this to what we have today. Based on my interpretation of the words, my impression is that "proposed" is a PR, same as today, and this doesn't change anything of the mechanics of the process other than the folder that the merged PR lands in and the status graphic after merging. Where we spend all the time in talking about a HIPE is as a PR, same as now.  Is that right?

dhh1128: My intent was that "proposed" would be a much shorter phase. As soon as the community gets to "this is more or less on the right track, with details to be worked out" it moves to "accepted". Accepted means "standards-track", with implementation underway. It doesn't mean "done." There is no FCP before merging --only after, when we are wanting to move the status to "adopted".

dhh1128: @swcurran I added the following note. Does it help?

NOTE: contributors who used the HIPE process prior to April 2019 should
see the acceptance process substantially simplified under this approach.
The bar for acceptance is not perfect consensus and all issues resolved;
it's just general agreement that a doc is "close enough" that it makes
sense to put it on a standards track where it can be improved as
implementation teaches us what to tweak.

swcurran: That's a more substantial change that I agree with.

kdenhartog:
I'm in favor of these changes. I think it would be valuable to provide further delineation between the *accepted* and *adopted* status as well. The situation I'm not sure of how we would handle with this process are. What happens when someone creates a protocol that is implemented, but never intends to be standardized? E.g. the ridesharing protocol. Are we ok with leaving this in the accepted status so people can discover it still?

Additionally, for asynchronous communication (such as this right now) should we move that discussion to rocketchat? I think this is what we end up doing in many cases anyways, so it should be acceptable. Also, if we do this I'd suggest we create a Indy-HIPE rocketchannel to constrain discussions about HIPEs to that channel.

Thoughts?

dhh1128:
@kdenhartog Re "a protocol that's implemented but not yet intended as a standard": What if we created a new folder--neither /features nor /concepts (or /text | /design), but something like /shared? For HIPEs of this type, getting merged into /shared is trivial, and the status just stays "shared" until someone proposes that move over to a standards track?

swcurran:
I like that idea @dhh1128. That would give us a place to put message families that we can collaborate on in the short term such as we are doing for IIW.

README.md
```diff
@@ -1,242 +1,127 @@
...
-  - Create a folder under `text/` for your feature, using the chosen name.
-    Copy `0000-template.md` to `text/<your folder name>/README.md`.
+  - Pick a folder name for your HIPE. The name should consist of a zero-padded
+    4-digit number plus a descriptive name for the topic. Use lower-kebab-case
+    for the descriptive name. Get the HIPE number by finding the number
+    of [the most recent pull request](https://github.com/hyperledger/indy-hipe/pulls)
+    and incrementing by 1. For example, if the highest numbered PR is 125, your
+    HIPE number would be 126. You should end up with a folder name like
+    `0026-my-cool-feature`.
+  - Create the folder and copy `0000-template.md` to `text/<your folder name>/README.md`.
```
TelegramSam: `text` here should be `feature` or `concept`, right?

TelegramSam:
What of the existing accepted HIPEs? Can we just leave them where they are but deprecate new additions to `text`?

dhh1128:
@TelegramSam @swcurran Regarding the folder names: let's create just 1 new one: /shared (for docs we want the community to know about, but we're not yet trying to standardize; has a very low bar for merges). Let's leave the names of /text and /design alone, because it's not worth the disruption of broken URIs. I can update the HIPE text to use the old names, but still describe the simplified process.

When the-project-formerly-known-as-envoy spins up, I am anticipating that a new RFC repo for it will be created, and that much of our content here will move over there. When that happens, let's use the improved folder names.

Agreed?

kdenhartog:
What if we matched the names of the folders to the statuses of the HIPE?

dhh1128:
@kdenhartog I would love it if that would work--but it has a very big downside, which is that hyperlinks all break if a HIPE changes status.

TelegramSam:
> When the-project-formerly-known-as-envoy spins up, I am anticipating that a new RFC repo for it will be created, and that much of our content here will move over there. When that happens, let's use the improved folder names.
> 
> Agreed?

I agree.

TelegramSam:
I had a thought today. If we lose the PR list as the list of 'in process' HIPEs, how can we know which ones need attention or review? I think they will melt right into the background and we might loose review activity. @dhh1128 @swcurran

dhh1128:
@TelegramSam Excellent point. Maybe we introduce a tracking doc on the wiki? Or we tell people that their PR needs to include an update to a status doc that's at the root of the repo, that we refer to?

swcurran:
From my earlier shot at this - I think the proposer should take responsibility for driving the PR.  If they let it languish, that's OK, if they want it moved, that's OK too.  They should ask for time on the Indy Agent call for it, and should lobby for co-sponsors.  The community still has the final vote and shows introduce.

Further, they can still submit PRs that can be discussed as PRs in evolving the HIPE.  The nice part is others can as well, without having to do a PR of a PR.

For the status, maybe we can write a script to scan the HIPEs (and PRs) and then gen a markdown file that can be posted somewhere - ideally posted automagically, but worst case as a PR to a readme file in the repo.

kdenhartog:
I'm in favor of @swcurran suggestions. This reinforces the doacracy process in my eyes, and would therefore help us to better prioritize work as a community.

swcurran:
Also remember that status changes are PRs, so that becomes the FCP process trigger.

TelegramSam:
@swcurran I like a script that outputs a directory. I wonder if the read the docs stuff is capable of that?

PR #115, opened 2019-04-11
avoid unneccessary base64 encode of protected data for tag computation

lovesh:
Signed-off-by: lovesh harchandani <lovesh.bond@gmail.com>


text/0028-wire-message-format/README.md
```diff
@@ -252,12 +252,12 @@ The protected data decodes to this:
...
-1. generate a content encryption key (symmetrical encryption key)
-2. encrypt the CEK for each recipient's public key using Anoncrypt (steps below)
+1. generate a content encryption key (symmetrical encryption key) CEK
+2. encrypt the CEK for each recipient's public key using Anoncrypt (steps below) and create a JSON array
     1. set `encrypted_key` value to base64URLencode(libsodium.crypto_box_seal(their_vk, cek))
         * Note it this step we're encrypting the cek, so it can be decrypted by the recipient
-3. base64URLencode the `protected` value
-4. encrypt the message using libsodium.crypto_aead_chacha20poly1305_ietf_encrypt_detached(message, protected_value_encoded, iv, cek) this is the ciphertext.
+3. set `protected` to base64 of the JSON array from step 2.
+4. encrypt the message using libsodium.crypto_aead_chacha20poly1305_ietf_encrypt_detached(message, JSON from step 2, iv, cek) this is the ciphertext.
```
kdenhartog: If I'm reading this properly, your proposal is to pass the bytes of JSON into the AAD? If so I believe this will make the tag unverifiable because JSON doesn't have deterministically serialize and deserialize, which I believe is necessary for the AAD operation. Am I reading this wrong?

kdenhartog:
I think this PR should be closed. It moves us away from the JWE standard and I don't believe it garnered enough efficiency worth accepting it. I believe this is why @lovesh closed the PR that referenced this.

dhh1128:
This has been superseded by a slightly modified technique that uses out-of-the-box JWE/JWS features.

PR #116, opened 2019-04-12
Transport Return Route

TelegramSam:
describes how a message over a transport would signal that the connection serves as an acceptable return route. This helps client only agents, including mobile apps.


text/return-route/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Tutorial
+[tutorial]: #tutorial
+
+When you send a message through a connection, you can use the `~transport` decorator on the message and specify `return_route`.
+
+```json
+{
+    "~transport": {
+        "return_route": "all"
```
dhh1128: What does "all" mean, and what are the other possible values of "return_route"?

dhh1128: Presumably, this means that the agent receiving the decorated message should attempt to send responses back over the same channel?

text/return-route/README.md
```diff
@@ -0,0 +1,92 @@
...
+    }
+}
+```
+
+Outbound messages can return the `~transport` decorator as well, with a `queued-message-count` attribute. This is useful for HTTP transports which can only receive one return message at a time.
```
dhh1128: We need a lot more detail here. If such a decorator is seen, how does the queued-message-count get used? Does the client call in with <message type x that asks for everything queued>, and get a response that bundles the whole queue or some subset thereof?

dhh1128: "queued-message-count" should be snake_case

text/return-route/README.md
```diff
@@ -0,0 +1,92 @@
...
+```
+
+If transport decorators are desired but no message needs to be sent, a `noop` message can be sent. This is useful if making a request to an agent just to establish a return route.
+
+```json
+{
+    "@type": "?/1.0/noop",
+    "~transport": {
+        "return_route": "thread",
+        "return_route_thread": "1234567899876543"
```
dhh1128: I don't get what "return_route_thread" means, or why it would be different from a normal @id that establishes a thread. Is this allowing us to change the return route for a thread that already exists?

text/return-route/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Reference
+
+[reference]: #reference
+
+`return_route` has the following acceptable values:
+
+- `none`: Default. No messages should be returned over this connection.
+- `all`: Send all messages for this key over the connection.
+- `thread`: Send all messages matching the key and thread specified in the `return_route_thread` attribute.
```
dhh1128: What "key" is specified in the "return_route_thread" attribute? I only see a thread id.

dhh1128:
Sam, this doc feels like a very promising one to me, and I'm glad it's in process. I think it should replace the ~please_duplex HIPE that I wrote several months ago (https://github.com/hyperledger/indy-hipe/blob/08927ab49754698de671862a2742ff486d8b2d06/text/please-duplex/README.md), because this is far simpler. However, there are several concepts from that HIPE draft that need to be discussed here. For example, what happens if someone doesn't want to honor the request for a return route? Is it illegal to use a different channel? Should an error or a warning be returned, and should it flow back on the channel over which the decorator arrived? What happens if the route is complex (flows through mediators and relays; maybe uses more than one transport) rather than simple--can we decorate forward messages with the decorator to ask intermediaries to support the routing, or is that out of scope? How can the sender of this decorator express the relative importance of the request, from their perspective?

text/return-route/README.md
TelegramSam: I meant `cryptographic`. added.

text/return-route/README.md
TelegramSam: This means that only messages with a matching thread ID would be returned. This allows the agent to only receive response messages to something they sent, and not messages of a type they might be unprepared to process at that moment.

text/return-route/README.md
TelegramSam: fixed. (this was not consistent with the rest of the doc)

TelegramSam: added additional commentary.

text/return-route/README.md
TelegramSam: added pointer to the reference section.

TelegramSam:
> What happens if the route is complex (flows through mediators and relays; maybe uses more than one transport) rather than simple--can we decorate forward messages with the decorator to ask intermediaries to support the routing, or is that out of scope? 

This question is interesting. As the `~transport` decorator is the first decorator not related to the specific message, this is new territory. As the transport between a mediator and the next hop is unknown to the original sender, I think this decorator should only be used on the outermost message, as that is the only transport that the original sender is aware of.

tplooker:
@TelegramSam @dhh1128  I like this HIPE however it has some implications around how agents will handle and manage messages. The implication this HIPE has is that all agents, if they have a message to send another agent and fail to find an endpoint in which to dispatch the message (or the message fails to transmit with the provided endpoint), it then must put in a que for that agent? If that is the case I think we need to have a broader discussion about this behavior before this progresses?

swcurran:
@TelegramSam - I think this needs more about the use case.  I think I get it, but it's really fuzzy.  A detailed example of how it would be used for a mobile and cloud agent to message back and forth would be a big help.

TelegramSam:
superseded by https://github.com/hyperledger/aries-rfcs/tree/master/features/0092-transport-return-route

PR #117, opened 2019-04-15
Rich schemas

brentzundel:
A HIPE that describes rich schema objects for verifiable credentials at a high level.


brentzundel:
If anyone could provide guidance on precisely why the DCO check is failing, I would appreciate it. As far as I can tell, all commits have been signed.

brentzundel:
replaced by PR #119

PR #120, opened 2019-04-19
propose HIPE: Message Trust Contexts

dhh1128:

### DIFF
--- a/text/message-trust-contexts/README.md
+++ b/text/message-trust-contexts/README.md
@@ -1,3 +1,7 @@
+[![moved to github.com/hyperledger/aries-rfcs repo](https://i.ibb.co/tBnfz6N/Screen-Shot-2019-05-21-at-2-07-33-PM.png)](https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0029-message-trust-contexts/README.md)
+
+New location: [aries-rfcs/concepts/0029-message-trust-contexts](https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0029-message-trust-contexts/README.md)
+
 # Message Trust Contexts
 
 ## Summary
--- a/text/message-trust-contexts/README.md
+++ b/text/message-trust-contexts/README.md
@@ -112,3 +112,9 @@ As message processing happens, the MTC isn't just updated. It should constantly 
 
 Part of the intention with the terse MTC notation is that conversations about agent trust should be easy and interoperable. When agents send one another [`problem-report` messages](https://github.com/hyperledger/indy-hipe/pull/65), they can turn MTCs into human-friendly text, but also use this notation: "Unable to accept a payment from message that lacks Integrity guarantees (-i)." This notation can help diagnose trust problems in logs. It may also be helpful with [message tracing](https://github.com/hyperledger/indy-hipe/pull/60), [protocol discovery](https://github.com/hyperledger/indy-hipe/pull/73), and agent testing.
 
+## Reference
+
+A complete reference implementation of MTCs in python is attached to this HIPE (see [mtc.py](mtc.py)).
+It could easily be extended with custom trust dimensions, and it would be simple to port to other
+programming languages. Note that the implementation includes unit tests written in pytest style,
+and has only been tested on python 3.x.
--- a/text/message-trust-contexts/README.md
+++ b/text/message-trust-contexts/README.md
@@ -1,0 +1,114 @@
+# Message Trust Contexts
+
+## Summary
+
+Introduces the concept of __Message Trust Contexts__ and describes how they are populated and used.
+
+## Motivation
+
+An important aim of DID Communication is to let parties achieve high trust. Such trust is vital in cases where money changes hands and identity is at stake. However, sometimes lower trust is fine; playing tic-tac-toe ought to be safe through agents, even with a malicious stranger.
+
+We may intuitively understand the differences in these situations, but intuition isn't the best guide when designing a secure ecosystem. Trust is a complex, multidimensional phenomenon. We need a formal way to analyze it, and to test its suitability in particular circumstances.
+
+## Tutorial
+
+_When Alice sends a message to Bob, how much should Bob trust it?_
+
+This is not a binary question, with possible answers of "completely" or "not at all". Rather, it is a nuanced question that should consider many factors. Some clarifying questions might include:
+
+* How much does Bob know about Alice? And what evidence is behind that knowledge?
+
+* Was the confidentiality and integrity of the message protected while it was in transit? If so:
+
+    * How strong is the encryption algorithm?
+    * How good are the protections on the keys?
+    * Is [perfect forward secrecy](https://en.wikipedia.org/wiki/Forward_secrecy) being used?
+
+* Does Bob have any kind of prior relationship to Alice, such that history increases (or decreases) trust?
+
+* Is the message [repudiable](https://github.com/hyperledger/indy-hipe/blob/master/text/0037-repudiation/README.md), or is Alice speaking "on the record"?
+
+* Could Alice use a malformed message to attack Bob (e.g., with buffer overflows, numeric range errors, unexpected fields, and so forth)?
+
+### Message Trust Contexts
+
+The DID Communication ecosystem formalizes the idea of a __Message Trust Context__ (__MTC__) to expose such questions, make their answers explicit, and encourage thoughtful choices based on the answers.
+
+An MTC is an object that holds trust context for a message. This context follows a message throughout its processing journey inside the agent that receives it, and it should be analyzed and updated for decision-making purposes throughout.
+
+Protocols should be designed with standard MTCs in mind. Thus, it is desirable that all implementations share common names for certain concepts, so we can discuss them conveniently in design docs, error messages, logs, and so forth. The standard dimensions of trust tracked in an MTC break down into two groups:
+
+#### Crypto-related
+* __Confidentiality__: Could the plaintext content of the message I just received over this channel have been observed by another party?
+* __Integrity__: Could the message have been tampered with since it was sent?
+* __Authenticated Origin__: Can I know the identity of the original sender with confidence? And if so, on what basis--were multiple factors of auth required to use the keys that encrypted the message? Were biometrics at play?
+* __Nonrepudiation__: Can I prove the identity of the sender to a third party? Note that this guarantee may or may not be a desirable quality.
+* __PFS__: Is perfect forward secrecy active?
+* __Uniqueness__: Can I know that this message is not being sent as part of a replay attack?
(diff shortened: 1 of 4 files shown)


text/message-trust-contexts/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Motivation
+
+An important aim of DID Communication is to let parties achieve high trust. Such trust is vital in cases where money changes hands and identity is at stake. However, sometimes lower trust is fine; playing tic-tac-toe ought to be safe through agents, even with a malicious stranger.
+
+We may intuitively understand the differences in these situations, but intuition isn't the best guide when designing a secure ecosystem. Trust is a complex, multidimensional phenomenon. We need a formal way to analyze it, and to test its suitability in particular circumstances.
+
+## Tutorial
+
+_When Alice sends a message to Bob, how much should Bob trust it?_
```
kdenhartog: I would suggest we change this to "When Alice's agent sends a message to Bob's agent, how much should Bob's agent trust it?"

Alice and Bob are not the ones making the determination of trust, but rather it is their agents that are. This is similar to the idea of the green lock in browsers where, as a user I can see that my browser considers the certificate valid and trusted, but I am not choosing to manually validate the keys and verify root authority. Rather I'm choosing to trust my browser is checking it for me properly and by extension I'm trusting my browser and not the keys. This is an important distinction that we need to make I believe.

text/message-trust-contexts/README.md
```diff
@@ -0,0 +1,120 @@
+# Message Trust Contexts
```
kdenhartog: While I like the mental model that is evoked by "Message Trust Contexts" I think "Message Security Contexts" more accurately portrays the features described in this HIPE. The main difference in my mind is that trust is a Human-to-Human or Human-to-Machine adjective in my vocabulary. In this HIPE we're describing features of Machine-to-Machine verification though which is why I'd prefer "Security" over "Trust".

text/message-trust-contexts/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+An important aim of DID Communication is to let parties achieve high trust. Such trust is vital in cases where money changes hands and identity is at stake. However, sometimes lower trust is fine; playing tic-tac-toe ought to be safe through agents, even with a malicious stranger.
+
+We may intuitively understand the differences in these situations, but intuition isn't the best guide when designing a secure ecosystem. Trust is a complex, multidimensional phenomenon. We need a formal way to analyze it, and to test its suitability in particular circumstances.
+
+## Tutorial
+
+_When Alice sends a message to Bob, how much should Bob trust it?_
+
+This is not a binary question, with possible answers of "completely" or "not at all". Rather, it is a nuanced question that should consider many factors. Some clarifying questions might include:
```
kdenhartog: This is the core essence of this work from what I can tell. I think this should be moved into the summary section with a bit of rephrasing. Something like: "Security of A2A is not binary and we need a method to describe the security requirements and verification methods of an A2A message. Message Security Contexts gives us a methodology to represent the spectrum of nuanced questions that should be considered when sending A2A messages securely."

text/message-trust-contexts/README.md
dhh1128: Good point. Will update.

text/message-trust-contexts/README.md
dhh1128: While I don't specifically disagree, I'd like to know whether your view of the security-vs-trust distinction resonates for some other community members. @TelegramSam or @swcurran?

text/message-trust-contexts/README.md
dhh1128: Okay, I'll update unless you raise a PR against my PR first. :-)

text/message-trust-contexts/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Custom Trust
+
+Specific agents may make trust distinctions that are helpful in their own problem domains. For example, some agents may evaluate trust on the physical location or IP address of a sender, or on the time of day that a message arrives. Others may use DIDComm for
+internal processes that have unique trust requirements over and above those that matter in interoperable scenarios, such as whether a message emanates from a machine running endpoint compliance software, or whether it has passed through intrusion detection or data loss prevention filters.
+
+Agent implementations are encouraged to add their own trust dimensions to their own implementations of a Message Trust Context, as long as they do not redefine the standard labels. In cases where custom trust types introduce ambiguity with trust labels, MTC notation requires enough letters to disambiguate labels. So if a complex custom MTC has fields named `intrusion_detect_ok`, `ipaddr_ok` (which both start like the standard `integrity`), and `endpoint_compliance` (which has no ambiguity with a standard token) it might be notated as:
+
+    mtc: +c+a+inte+intr+ip-n-p-e
```
kdenhartog: As a part of verifying MSC we should have the code that is verifying the security also report if the messages actually contain what is claimed. For example, the pack and unpack method should verify the Authenticated origin (used Authcrypt) and then pass a response to the MTC verifier. This follows the "trust but verify" principle and helps us avoid an attack where a claim is made for some context, but is not actually performed.

text/message-trust-contexts/README.md
kdenhartog: I can submit this as a PR to your fork.

cywolf:
I'm looking at how these flags would apply within the context of an Indy agent with pairwise connections. It might be helpful to have a summary of the identifiers in the HIPE, I borrowed these from the python code.

`SIZE_OK`: The binary message isn't too large.
`DESERIALIZE_OK`: It's JSON.
`KEYS_OK`, `VALUES_OK`: It matches an expected schema.

These flags seem to apply to loading and decrypting a wire message, and then again to validating the contents as a known agent message. They also don't matter at all to the message handlers, as the handler wouldn't be called if any of these failed. In terms of what might be returned in a problem report, something like `message trust context failed: -s` seems a lot more opaque than just `message size too large`, so I can't say I really see the value in tracking these.

`CONFIDENTIALITY`: Presumably true for an encrypted message. At the agent level I may not have enough information about the transports used (TLS etc) to say whether an unencrypted message was confidential. Even if my incoming transports are secure I don't know that the sender hasn't forwarded it through unsecure transports.

`INTEGRITY`: Mostly the same context as `CONFIDENTIALITY`, I believe. I suppose a MITM could substitute an entirely different message if they intercepted an anon-crypted one, but without knowing what the original contained.

`AUTHENTICATED_ORIGIN`: If the message is auth-crypted then I can associate it with a connection and whatever trust I've assigned to that. As to "were multiple factors of auth required to use the keys that encrypted the message? Were biometrics at play?" – I don't know how that could be attested in a trustable way, and that seems to imply multiple levels of authentication instead of a yes/no value.

`NONREPUDIATION`: I think this would require a wire message format that also included signatures? Following the signing HIPE, only portions of the agent message can be assigned signatures, not the message itself.

`PFS`: This might apply at the transport layer or it might not, I don't think the agent is qualified to say. Maybe if there were an additional layer of automatic key rotation built in?

`UNIQUENESS`: If the agent stored hashes of every incoming message that might be enough to determine this, but that seems like a trade-off that probably isn't worth implementing in most cases.

`LIMITED_SCOPE`: It looks like this one the agent can determine by whether the recipient key used in the wire message is one from the agent's public DID Document. But in that case the message wouldn't be associated with a pairwise connection, so it would have limited trust anyway.

I can see `PFS` and `UNIQUENESS` being optionally enforced at the agent level for all pairwise communication, but they seem unlikely to be selectively required at the handler level. Otherwise, message handlers only seem to vary in whether they require an auth-crypted message associated with an existing pairwise connection. Most do, and the rest would have specific conditions under which they would accept the message, like if it's associated with a connection invitation. I could see non-repudiation being important in some message handlers in some contexts but I'm not sure how that would be implemented at the moment.

dhh1128:
@cywolf Thank you for the detailed feedback!

I'd like to explore some nuances related to your comments; depending on how we settle them in the comment thread, we may need to update the HIPE to simplify, or to clarify why the complexity is possible.

Regarding DESERIALIZE_OK equating to "It's JSON": I deliberately didn't name this "JSON_OK" because I could imagine CBOR or other technologies being used in some future version of DIDComm, and I didn't want the JSON assumption to be overly strong. Also, it is possible for something to be JSON but to not deserialize OK, as with [JSON that repeats a key](https://stackoverflow.com/a/21833017/2953158).

I agree that these attributes of a message trust context are most likely to be relevant early in the processing cycle (e.g., when turning a wire message into plaintext). However, I'm not sure the division of responsibilities will always be perfectly crisp; for example, I could imagine a handler evaluating VALUES_OK. I could even imagine a handler being implemented as a derived class hierarchy, where a base class does some of the evaluation, and derived classes do specialized validation for specific versions of the protocol. I also don't think we can assume that message trust context evaluation will be purely sequential; I could imagine multiple forms of validation happening in parallel (e.g., on different threads, in different classes).

You are correct that CONFIDENTIALITY is generally true if a message is encrypted. We are in the habit of assuming that all messages are encrypted in DIDComm (whether authcrypted or anoncrypted)--but we already have at least one example where this is not true--the connection invitation. Other low-security protocols (e.g., tictactoe) might be usable over an insecure channel as well. And Tobias has proposed non-confidential but signed operations as a general mode of operation for connectionless ("ephemeral") interactions. So we can't assume that confidentiality will always be true. When it is not true because of encryption, then you are correct that we could impute it based on transport--but only if we know with certainty the entire transmission pipeline.

If you have CONFIDENTIALITY, you probably also have INTEGRITY. But this is not guaranteed, and in fact certain crypto algorithms are susceptible to chosen ciphertext attacks exactly because the two are not identical. You can also have integrity without confidentiality, as when a message is signed but plaintext.

You are right that if AUTHENTICATED_ORIGIN considers the issue of biometrics, it's no longer binary. So the HIPE is inconsistent on this point, and needs to be fixed to either remove the mention of biometrics and keep the flag binary, or to propose some way to deal with non-binary-ness. Authcrypted messages give us authenticated origin, but there might be additional ways to achieve this goal, so I wouldn't make the equivalence between this flag and the authcrypt mechanism overly strong.

NONREPUDIATION: yes, only possible if whole message become signed. Otherwise the flag would be false.

PFS is the subject of indy HIPE 0024; we would not depend on transport to provide it. It should be possible to set this flag without knowledge of transport. If the wire format switches to use MLS (as @kdenhartog has talked about), PFS remains transport-independent.

UNIQUENESS was intended to just ask if the message has a nonce or ID that should guard against replay attack. I think most of our messages will automatically be checked for unique ID since we will have to look up thread status as we process--so I'm not sure this adds any overhead.

LIMITED_SCOPE is just a way to gauge privacy based on whether the DID is a public one or is n/pair-wise. The combination of LIMITED_SCOPE=false and NONREPUDIATION=true has ramifications that agents should understand.

dhh1128:
This is superseded by Aries RFC 0029.

PR #152, opened 2020-01-22
UPDATE: Rich Schema HIPE 

brentzundel:
update schema examples and methods
Signed-off-by: Brent <brent.zundel@gmail.com>


text/0149-rich-schema-schema/README.md
```diff
@@ -237,102 +240,46 @@ This will be done following the pattern for `context_handler.py` and
...
-                    "skos": "http://www.w3.org/2004/02/skos/core#",
-                    "snomed": "http://purl.bioontology.org/ontology/SNOMEDCT/",
-                    "void": "http://rdfs.org/ns/void#",
-                    "xsd": "http://www.w3.org/2001/XMLSchema#",
-                    "xsd1": "hhttp://www.w3.org/2001/XMLSchema#"
+            "@context": [
+                "https://www.w3.org/ns/did/v1", 
+                "did:sov:yfXPxeoBtpQABpBoyMuYYGx"
+            ],
+            "id": "7u7cVZWrQ5VTdJxAsaaGFGqaDuuS4GU73d8DNWzVuMSX",
```
ashcherbakov: Should it have `did:sov` prefix?

ashcherbakov:
Thanks for sending the PR with update.

I think most of comments/questions from https://github.com/hyperledger/indy-hipe/pull/151 are applied here as well. 

Also I believe it would be great to decide the following:

1. How the @id's DID id-string is created, do we have any assumptions there or it can be any valid DID.
1. How a Rich Schema object is stored in the ledger State
1. What is a general format of Rich Schema-related objects, what are common fields
    - How we split data (object itself) and metadata
    - What metadata fields do we need (do we need type, name, version etc.?)
    - Do we store the data (object itself) as a JSON string, or as a dict within the parent JSON request/transaction.  
    - Why do we need a @context object as a metadata field, and as part of Schema object itself.

brentzundel:
this was made part of PR #156 
closing

PR #154, opened 2020-02-10
Propose HIPE: Rich Schema Encodings

brentzundel:
Describes Indy Node and other methods for adding rich schema encoding objects to Indy.
This is part of the Rich Schema work described [here](https://github.com/hyperledger/aries-rfcs/tree/master/concepts/0250-rich-schemas).


brentzundel:
this was made part of PR #156 
closing

PR #155, opened 2020-02-18
update HIPE 120: modified relationship image to clarify object references

brentzundel:
Signed-off-by: Brent <brent.zundel@gmail.com>


text/0120-rich-schemas-common/relationship-diagram.puml
```diff
@@ -13,12 +13,22 @@ object "Verifiable credential" as verif_cred {
...
+
+pres "<b>1" --> "<b>0..n" pres_def
+pres "<b>1..n" --> "<b>0..n" verif_cred
+
+pres_def "\n<b>0..n" --> "<b>0..n" rich_schema
+pres_def "<b>0..n" --> "<b>0..n" map
+pres_def "<b>0..n" --> "<b>0..n" cred_def
+
+verif_cred "<b>1" --> "<b>0..n" cred_def
+verif_cred "<b>1..n" --> "<b>0..n" rich_schema
```
ashcherbakov: I think this relationship is implicit. A verifiable credential references a CredDef which in turn references Schemas (as well as Encodings) via Mapping. 
So, maybe it's sufficient to have explicit relationship between verif_cred and cred_def only?

text/0120-rich-schemas-common/relationship-diagram.puml
```diff
@@ -13,12 +13,22 @@ object "Verifiable credential" as verif_cred {
...
-pres_def "<b>0..n" -- "<b>0..n" rich_schema
-pres_def "\n<b>0..n" -- "<b>0..n\n" cred_def
-verif_cred "<b>1" -- "<b>0..n" cred_def
-map "\n<b>1..n" -- "<b>0..n" enc
+
+pres "<b>1" --> "<b>0..n" pres_def
+pres "<b>1..n" --> "<b>0..n" verif_cred
+
+pres_def "\n<b>0..n" --> "<b>0..n" rich_schema
+pres_def "<b>0..n" --> "<b>0..n" map
```
ashcherbakov: Is it implicit or explicit relationship? In other words, can PresentationDefinition reference a Mapping without a CredDef, or it references a Mapping implicitly via CredDef only?

text/0120-rich-schemas-common/relationship-diagram.puml
brentzundel: We agree that this should be sufficient.
The relationship needs to be explicit in the credential in order to be compliant with the W3C VC Data Model. We anticipate the schema to be provided as part of the `type` property: https://www.w3.org/TR/vc-data-model/#types. This allows JSON-LD processors to handle the different claims included in the credential.
The cred def, on the other hand, is provided as part of the `credentialSchema` property: https://www.w3.org/TR/vc-data-model/#data-schemas and https://www.w3.org/TR/vc-data-model/#zero-knowledge-proofs. This allows ZKP-style VCs to be cryptographically verifiable.

text/0120-rich-schemas-common/relationship-diagram.puml
brentzundel: I think it should be explicit, but the presentation definition is still being developed.
At this point, the presentation definition may specify the desired cred def, and thereby the desired mapping and schemas, or it may more generally specify a particular mapping (we anticipate that mappings will be re-used between issuers). Or a presentation definition may be even more general and specify only the schema properties that are desired. 

It may be helpful to think of this as a spectrum of possible specificity. For each attribute a verifier desires to see, the verifier can specify the acceptable range of possible sources:
1. At the least restrictive, the verifier needs to specify the schema which defines the property whose value is the desired attribute.
1. At the next level, the verifier might require that an attribute be supplied via a certain mapping (which includes the schema and context information, but does not require a certain cred def or issuer). 
1. The next level would be for the verifier to require a certain issuer (or set of issuers), each of which includes a possible set of cred defs, each of which points to a mapping. 
1. Finally, the verifier may specify a particular cred def, which includes the issuer, mapping, schemas, and contexts which are desired.

At a minimum, the schema needs to be included
I hope that clarifies things.

text/0120-rich-schemas-common/relationship-diagram.puml
ashcherbakov: Ok, thanks for the explanation. BTW are we OK with the naming here?
It looks like Credential's `type` points to a Schema, and Credential's `credentialSchema` points to a CredDef. This can be a bit confusing...

brentzundel:
@dhh1128 @nage this should be ready to merge.

brentzundel:
the naming isn't great, but we can't change the credential property names
(they are named in the spec), the schema is a schema (so I'm not sure what
else we would call it), and I'm not sure we want to change CredDef.

PR #159, opened 2020-05-19
Use correct license

ryjones:


ryjones:
@swinslow - thoughts?

swinslow:
Hmmm... @dhh1128 good point given that the past contributions to this repo were originally made under Apache-2.0.

In license scans I had flagged this to @ryjones because the [Hyperledger charter](https://www.hyperledger.org/about/charter) says that Hyperledger documentation should be under CC-BY-4.0 (see section 13(c)). But it likely isn't as straightforward as just changing the license since the contributions to date were made under Apache-2.0.

@ryjones let me look into this a bit further and I'll circle back with you -- suggest holding off on the PR for now. Apologies for causing any confusion here.

ryjones:
This will never be resolved.

PR #161, opened 2020-09-29
Document the findings about the use of intervals in Proof Requests and calls to indy-node to get ledger information

swcurran:
A PR to capture the learning we did last week about Intervals in revocation requests.  This documents what the community agreed to on the Aries WG Call and provides practical guidance to Prover and Verifier builders.


dhh1128 (approved):
discussed on Aries community call and on Indy chat channels; LGTM

PR #162, opened 2021-01-23
WIP: First draft of HIPE for plugin helpers

esplinr:
Introduces features previously discussed in the Indy Contributors calls: frozen ledgers and default fee handlers.


text/0162-plugin-removal-helpers/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Frozen Ledgers
+[frozen-ledgers]: #frozen-ledgers
+
+### Tutorial
+[frozen-ledgers-tutorial]: #frozen-ledgers-tutorial
+
+If a ledger has been created, but will never be used, it can be frozen. The LEDGERS_FREEZE transaction will record the root hashes of one or more ledgers, and perpetually use those hashes when computing audit ledger validity. This has two important advantages:
+
+* It preserves the ability to do rolling updates. If an update removes a plugin which includes custom ledgers, consensus would be interrupted and new transactions would not be able to be ordered until all nodes completed the update. This is because updated nodes would not be able to process new transactions that were ordered by the non-updated nodes. But if we freeze all ledgers associated with the plugin, we can do rolling updates because new transactions related to the plugin will be impossible.
```
askolesov, line 37:
```suggestion
* It preserves the ability to do rolling updates. If an update removes a plugin which includes custom ledgers, consensus would be interrupted in case of new plugin related transactions during a rolling update. This is because updated nodes would not be able to process new transactions that were ordered by the non-updated nodes. But if we freeze all ledgers associated with the plugin, we can do rolling updates because new transactions related to the plugin will be impossible.
```

text/0162-plugin-removal-helpers/README.md
```diff
@@ -0,0 +1,155 @@
...
+* Execute a migration script to drop the ledgers.
+
+Permissions around the LEDGERS_FREEZE transaction are managed in the auth_rules with a default permission requiring the approval of three trustees ([the same default auth_rule for upgrading the ledger](https://github.com/hyperledger/indy-node/blob/master/docs/source/auth_rules.md)).
+
+### Reference
+[frozen-ledgers-reference]: #frozen-ledgers-reference
+
+If a ledger is frozen it can be used neither for reading nor for writing. It will not be caught up by new nodes and can be safely removed. Default ledgers such as the domain, config, pool, and audit ledgers cannot be frozen.
+
+Implementation notes:
```
askolesov, line 53:
```suggestion
#### Implementation notes:
```

text/0162-plugin-removal-helpers/README.md
```diff
@@ -0,0 +1,155 @@
...
+* Implementation of this feature is possible because the catchup of the config ledger happens before catchup of other ledgers.
+* We were nervous that replacing a ledger with a frozen ledger hash would break foreign keys in auth_rules, but our testing showed that this is not an issue.
+* Frozen ledgers cannot be unfrozen, but dropped ledgers can be recreated with a different ledger ID.
+
+Implementation steps:
```
askolesov, line 58:
```suggestion
#### Implementation steps:
```

text/0162-plugin-removal-helpers/README.md
```diff
@@ -0,0 +1,155 @@
...
+Permissions around the LEDGERS_FREEZE transaction are managed in the auth_rules with a default permission requiring the approval of three trustees ([the same default auth_rule for upgrading the ledger](https://github.com/hyperledger/indy-node/blob/master/docs/source/auth_rules.md)).
+
+### Reference
+[frozen-ledgers-reference]: #frozen-ledgers-reference
+
+If a ledger is frozen it can be used neither for reading nor for writing. It will not be caught up by new nodes and can be safely removed. Default ledgers such as the domain, config, pool, and audit ledgers cannot be frozen.
+
+Implementation notes:
+* Implementation of this feature is possible because the catchup of the config ledger happens before catchup of other ledgers.
+* We were nervous that replacing a ledger with a frozen ledger hash would break foreign keys in auth_rules, but our testing showed that this is not an issue.
```
askolesov, line 55:
```suggestion
* We were nervous that removing transactions would break foreign keys in auth_rules, but our testing showed that this is not an issue.
```

text/0162-plugin-removal-helpers/README.md
```diff
@@ -0,0 +1,155 @@
...
+* Frozen ledgers cannot be unfrozen, but dropped ledgers can be recreated with a different ledger ID.
+
+Implementation steps:
+1. Implement LEDGERS_FREEZE transaction handler that will write information about frozen ledger to state
+2. Implement GET_FROZEN_LEDGERS transaction handler that will return information about frozen ledgers from state
+3. Add dynamic validation to all handlers: check if the ledger is frozen
+4. Update audit logic by taking into account the root hashes of frozen ledgers
+5. Fix catch up to not catch up frozen ledgers **TODO: mirror requests or not?**
```
askolesov, line 63:
```suggestion
5. Fix catch up to not catch up frozen ledgers.
```

We've figured out that it's not necessary.

text/0162-plugin-removal-helpers/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+Return frozen ledgers and their root hashes.
+```
+GET_FROZEN_LEDGERS
+```
+
+### Drawbacks
+[frozen-ledgers-drawbacks]: #frozen-ledgers-drawbacks
+
+The LEDGERS_FREEZE transaction allows removal of the plugin that created the ledger if the ledger was never used and the hash on the audit ledger never changed. Otherwise, the plugin will need to remain installed to allow historical transactions to replayed with the correct ledger hash.
```
askolesov, line 96:
```suggestion
The LEDGERS_FREEZE transaction allows removal of the plugin that created the ledger if the ledger was never used and the hash on the audit ledger never changed. Otherwise, the plugin will need to remain installed to allow historical transactions to be replayed with the correct ledger hash.
```

text/0162-plugin-removal-helpers/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+[Indy Authorization Rules](https://github.com/hyperledger/indy-node/blob/master/docs/source/requests.md#auth_rule) provide a flexible way of authorizing write transactions to the domain ledger. This authorization can [depend on the payment of fees tracked by ledger plugins](https://github.com/sovrin-foundation/libsovtoken/blob/master/doc/fees.md).
+
+Because Indy currently has no native concept of fees, if a plugin that implemented fees is removed, historical transactions can no longer be validated and will prevent new nodes from catching up on the ledger. In many cases it is sufficient to know that the historical transactions were valid at the time of writing. In these cases a default fee handler can approve the historical transactions.
+
+### Reference
+[default-fee-handler-reference]: #default-fee-handler-reference
+
+The default fee handler is implemented in a new class ...
+**TODO**
```
askolesov, line 128: The default fee handler will be a copy of sovtoken fee handler with the validation disabled.

Transaction handler consists of two parts: validation logic (static and dynamic) and business logic.

For new transactions validation is performed. During the catchup process, no validation is performed (because since the transaction is in the ledger it means that validation it was already performed and we don't need to do it twice).

So, if we copy the handler and disable validation, it will allow catchup but will not allow new transactions.

Business logic for fees related transactions must remain the same. Enabling of those transactions and the logic that will allow fees - up to a specific implementation.

Tokens are protected against attacks (even memory-related) because hackers have to perform the attack on a majority of nodes simultaneously (the property of blockchain).

text/0162-plugin-removal-helpers/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+### Drawbacks
+[default-fee-handler-drawbacks]: #default-fee-handler-drawbacks
+
+* Historically there were concerns about having payment functionality included in Hyperledger projects, but as distributed payments have become more mainstream, it now seems acceptable to include into Indy generic primitives related to payments.
+* The default fee handler will not perform the same validation on a historical transaction as was done in the removed plugin at the time that the transaction was originally completed. Therefore, it is most suitable for ledgers that do not make strong guarantees of historical validity, such as ledgers used for development and testing.
```
askolesov, line 134:
```suggestion
* The default fee handler will not perform validation at all. It will replay it in the same way as the plugin did.
```
> Therefore, it is most suitable for ledgers that do not make strong guarantees of historical validity, such as ledgers used for development and testing.

As far as I know, historical validity is the property of all ledgers.

text/0162-plugin-removal-helpers/README.md
```diff
@@ -0,0 +1,155 @@
...
+* If auth_rules with fees are defined, but no plugin is installed to process them, the rules will fall back to the default fee handler and authorization will be granted. This permissive behavior could be dangerous in some circumstances, and network administrators should ensure that auth_rules do not rely on fees when no plugin is installed to handle them. This is consistent with other Indy configuration options which are permissive by default to assist new users, but expected to be locked down for production use.
```
askolesov, line 135: > If auth_rules with fees are defined, but no plugin is installed to process them, the rules will fall back to the default fee handler and authorization will be granted.

Right, there will be no fee handler so rules will just be ignored. Just a note.

askolesov (commented):
Looks wonderful. Just a couple of minor suggestions.

text/0162-plugin-removal-helpers/README.md
esplinr: I added this to the document with a number of changes to try and make it easier to understand. Please confirm that I didn't mess it up.

text/0162-plugin-removal-helpers/README.md
esplinr: Excellent feedback. I reworded my comment, so hopefully it is now accurate.

text/0162-plugin-removal-helpers/README.md
esplinr: This prompted me to try and clarify that sentence. Let me know if it is better or worse.

text/0162-plugin-removal-helpers/README.md
```diff
@@ -0,0 +1,162 @@
...
+
+### Reference
+[default-fee-handler-reference]: #default-fee-handler-reference
+
+The default fee handler will be a copy of [the sovtoken fee handler (FeesAuthorizer)](https://github.com/sovrin-foundation/token-plugin/blob/master/sovtokenfees/sovtokenfees/fees_authorizer.py#L26) with the validation disabled.
+
+Any transaction handler consists of two parts: validation logic (static and dynamic) and business logic.
+
+For fee transactions, transaction validation will be performed on new transactions, but no validation will be performed during the catchup process. This is because the transaction is already on the ledger, so we know that validation was already performed and we don't need to do it again. By copying the handler and disabling validation, it will allow catchup but will not allow new transactions with fees.
```
askolesov, line 131:
```suggestion
Any transaction handler consists of two parts: validation logic (static and dynamic) and business logic. Validation is performed on new transactions, but no validation is performed during the catchup process. By copying the handler and disabling validation, it will allow catchup but will not allow new transactions with fees.
```

text/0162-plugin-removal-helpers/README.md
```diff
@@ -0,0 +1,162 @@
...
+
+Business logic for fee related transactions will continue to be determined by the auth_rules and logic implemented in a plugin which defines a payment implementation.
```
askolesov, line 133:
```suggestion
Business logic for fee related transactions will continue to be determined by the default handler. An implementation of plugin should unlock fee transactions and define a payment implementation.
```

text/0162-plugin-removal-helpers/README.md
askolesov: Your version looks much better. I've added a couple of suggestions how to make it even easier to understand (in the new review).

text/0162-plugin-removal-helpers/README.md
askolesov: Perfect

askolesov (approved):
Cool. I think the document is ready. Just a couple of optional suggestions from my side.

text/0162-plugin-removal-helpers/README.md
esplinr: I get it now. You are saying that every transaction handler skips validation on catch-up (it isn't specific to the fee handler). I didn't understand that before. I've updated the text to be more accurate. Thank you.

text/0162-plugin-removal-helpers/README.md
esplinr: Hopefully I got this right this time.

esplinr:
@askolesov : I'm concerned that I misunderstood this paragraph when I wrote it:

>  If auth_rules with fees are defined, but no plugin is installed to process them, the validation will fall back to the default fee handler which will ignore them and consequently grant authorization. This permissive behavior could be dangerous in some circumstances, and network administrators should ensure that auth_rules do not rely on fees when no plugin is installed to handle them. This is consistent with other Indy configuration options which are permissive by default to assist new users, but expected to be locked down for production use.

Since we also say:

>  The default fee handler will contain a copy of the sovtoken fee handler but with validation disabled, so it will allow catchup but will not allow new transactions with fees.

Doesn't that mean that the behavior is not permissive? Meaning all auth_rules that exclusively rely on fees for authorization will always fail unless their is a token plugin to override the default fee handler?

text/0162-plugin-removal-helpers/README.md
askolesov: Let let me dive into technical details a bit more. That's how I see this:

We have SET_FEE transaction and the corresponding handler in the plugin. Handler consists of two parts: validation logic and execution logic. After moving the handler to the node validation logic will be replaced with `throw NotAllowedException()`. The new plugin should:
1. Override default fee handler and replace validation logic with a real one. Execution logic must be kept the same, otherwise, catchup will be broken. That's what I mean by `unlocking fee transactions`. So new fee transactions will be allowed.
2. Implement token transactions like, for example, EMIT_TOKENS, TRANSFER_TOKENS and so on.
3. Implement the logic that will reduce the transaction's author balance if the transaction has a fee defined.

By `defining a payment implementation` I understand items 2 and 3.

askolesov:
@esplinr It's important to distinguish two cases:
1. Somone tries to execute a transaction that has a fee defined. There will be no logic to decrease the author's balance without plugin so the information that the transaction has fee will be simply ignored. And this is the permissive behaviour you are talking in the first part.
2. Someone tries to change fee for a transaction (i.e. sends SET_FEE transaction). This is what your second cite is about. In this case, the transaction will be rejected. So yes, it's not permissive behaviour.

text/0162-plugin-removal-helpers/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Default Fee Handlers
+[default-fee-handler]: #default-fee-handler
+
+### Tutorial
+[default-fee-handler-tutorial]: #default-fee-handler-tutorial
+
+[Indy Authorization Rules](https://github.com/hyperledger/indy-node/blob/master/docs/source/requests.md#auth_rule) provide a flexible way of authorizing write transactions to the domain ledger. This authorization can [depend on the payment of fees tracked by ledger plugins](https://github.com/sovrin-foundation/libsovtoken/blob/master/doc/fees.md).
+
+Because Indy currently has no native concept of fees, if a plugin that implemented fees is removed, historical transactions can no longer be validated and will prevent new nodes from catching up on the ledger. In many cases it is sufficient to know that the historical transactions were valid at the time of writing. In these cases a default fee handler can approve the historical transactions.
```
m00sey, line 122: I think "Because Indy currently has no native concept of fees" this should be a separate HIPE. Fees seem like a side effect of the plugin architecture although documented.

Looking toward the future I think having a defined "fee handler architecture" in Indy is valuable, allowing other to extend it but supporting the current work done by sovrin.

text/0162-plugin-removal-helpers/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Reference
+[default-fee-handler-reference]: #default-fee-handler-reference
+
+Implementation will require adding to Indy [the SET_FEE transaction copied from the sovtoken plugin](https://github.com/sovrin-foundation/token-plugin/blob/a635780361a7478ea4b2f47fcffee78b150fdea3/sovtokenfees/sovtokenfees/req_handlers/write_handlers/set_fees_handler.py#L20) and [a default fee handler copied from the sovtoken fee handler (FeesAuthorizer)](https://github.com/sovrin-foundation/token-plugin/blob/master/sovtokenfees/sovtokenfees/fees_authorizer.py#L26). The fee handler will have the validation logic replaced with `throw NotAllowedException()`.
```
m00sey, line 127: SET_FEE and FeesAuthorizer seem like an independent HIPE to support generic fees?

m00sey (commented):
Reading this I have an overarching feeling it's two HIPEs

one for ledger freezing, a QoL improvement for ledger plugin/development
and the second for the the introduction of fees at an Indy (vs plugin) level

I appreciate both are needed to resolve the current sovtoken issues, but feel they are worth discussing independently as I think they individually provide value

text/0162-plugin-removal-helpers/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+Removing the plugin has three important consequences:
+* The data stored on the custom ledger created by the plugin will be deleted.
+* The audit ledger will no longer be able to obtain the root hash of any custom ledgers from the plugin.
+* Any authorization rules (auth_rules) on the config ledger that define fees as being necessary to authorize writes to the domain ledger can no longer be handled by the plugin.
+
+The first consequence is intended. This HIPE proposes feature to address the remaining two consequences to ensure that the network remains functional after the plugin is removed.
```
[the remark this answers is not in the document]
esplinr: Good catch.

text/0162-plugin-removal-helpers/README.md
esplinr: I took your advice and split this into its own HIPE.

text/0162-plugin-removal-helpers/README.md
esplinr: I took your advice and split it into its own HIPE.

text/0162-plugin-removal-helpers/README.md
```diff
@@ -0,0 +1,160 @@
...
+* Send a LEDGERS_FREEZE transaction for the ledgers that you want to drop.
+* Upgrade all nodes to remove the plugin.
+* Execute a migration script to drop the ledgers.
+
+Permissions around the LEDGERS_FREEZE transaction are managed in the auth_rules with a default permission requiring the approval of three trustees ([the same default auth_rule for upgrading the ledger](https://github.com/hyperledger/indy-node/blob/master/docs/source/auth_rules.md)).
+
+### Reference
+[frozen-ledgers-reference]: #frozen-ledgers-reference
+
+If a ledger is frozen it can be used neither for reading nor for writing. It will not be caught up by new nodes and can be safely removed. Default ledgers such as the domain, config, pool, and audit ledgers cannot be frozen.
```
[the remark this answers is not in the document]
esplinr: Thoughts @Toktar ?

text/0162-plugin-removal-helpers/README.md
```diff
@@ -0,0 +1,160 @@
...
+## Frozen Ledgers
+[frozen-ledgers]: #frozen-ledgers
+
+### Tutorial
+[frozen-ledgers-tutorial]: #frozen-ledgers-tutorial
+
+If a ledger has been created, but will never be used, it can be frozen. The LEDGERS_FREEZE transaction will record the root hashes of one or more ledgers, and perpetually use those hashes when computing audit ledger validity. This has two important advantages:
+
+* It preserves the ability to do rolling updates. Without this capability, attempting to remove a plugin during a rolling update risks breaking consensus if a transaction is submitted to the ledger during the roll out period. This is because updated nodes would not be able to process new transactions that were ordered by the non-updated nodes. But if we freeze all ledgers associated with the plugin, we can do rolling updates because new transactions related to the plugin will be impossible.
+* It allows the removal of old plugins. Frozen ledgers can be removed and there will be no left over data from the removed plugin. The removal of old plugins will also help reduce the cost of keeping the network stable and secure.
```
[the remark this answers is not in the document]
esplinr: Yes, it exists in the freeze_ledgers transaction stored in the state tree (the state tree contains the most recent version of all transactions). This allows the hash to be used when the audit ledger is calculating consistency.

text/0162-plugin-removal-helpers/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+We include both features in the same HIPE because they share a common motivation and are likely to be used together.
+
+## Frozen Ledgers
+[frozen-ledgers]: #frozen-ledgers
+
+### Tutorial
+[frozen-ledgers-tutorial]: #frozen-ledgers-tutorial
+
+If a ledger has been created, but will never be used, it can be frozen. The LEDGERS_FREEZE transaction will record the root hashes of one or more ledgers, and perpetually use those hashes when computing audit ledger validity. This has two important advantages:
```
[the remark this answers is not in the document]
esplinr: This touches on two separate properties of our blockchain:
* The ability to achieve consensus when writing a new transaction to the ledger.
* The ability to audit the history of the ledger to prove that there was no tampering.

When a ledger is frozen, the root hash of the deprecated ledger is available to be used by the audit ledger when computing consensus. But we do not store the entire history of root hashes, so it does not help if someone tries to recompute a history that relies on the removed ledger. The proposed implementation of frozen ledgers only works for Sovrin because MainNet's token ledger was unused, and we don't guarantee historical reliability on StagingNet and BuilderNet.

I expanded on the explanation in the "Drawbacks" section to hopefully make that more clear.

esplinr:
For reviewers: a convenient link to the [pull request for the Sovrin SIP](https://github.com/sovrin-foundation/sovrin-sip/pull/28/) that is linked in the HIPE.

text/0162-plugin-removal-helpers/README.md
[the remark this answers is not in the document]
m00sey: I'll take a look, thank you.

text/0162-plugin-removal-helpers/README.md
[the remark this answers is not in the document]
Toktar: I added this error message for all client actions with a frozen ledger. But it's not a final solution and may be changed if you have suggestions.
```
InvalidClientRequest(request.identifier, request.reqId,
                                           "'{}' transaction is forbidden because of "
                                           "'{}' ledger is frozen".format(self.txn_type, self.ledger_id))
```

text/0162-frozen-ledgers/README.md
```diff
@@ -0,0 +1,129 @@
...
+2. Implement GET_FROZEN_LEDGERS transaction handler that will return information about frozen ledgers from state.
+3. Add dynamic validation to all handlers: check if the ledger is frozen.
+4. Update audit logic by taking into account the root hashes of frozen ledgers.
+5. Fix catch up to not catch up frozen ledgers.
+
+### State schema:
+There will be a key in CONFIG_LEDGER that will store a list of frozen ledgers in the following format:
+
+```
+2:FROZEN_LEDGERS = [
```
Toktar, line 68: Sorry for the fickleness. Can we change this to `4:FROZEN_LEDGERS`?
After moving this transaction from indy-node to indy-plenum, its marker has changed.

esplinr: Thank you for the correction.

text/0162-plugin-removal-helpers/README.md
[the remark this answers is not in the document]
esplinr: Thank you @Toktar . I added the error messages to the HIPE.

m00sey (approved):

esplinr:
PR's that implement the functionality described here:
https://github.com/hyperledger/indy-plenum/pull/1502
https://github.com/hyperledger/indy-node/pull/1641
https://github.com/hyperledger/indy-sdk/pull/2356

lynnbendixsen (approved):
This HIPE solves the issue in an acceptable way.

text/0162-frozen-ledgers/README.md
```diff
@@ -0,0 +1,139 @@
...
+## Summary
+[summary]: #summary
+
+Ledgers that are created by Indy plugins cannot be removed without breaking consensus. This HIPE describes functionality to freeze a ledger so that it can be safely removed.
+
+
+## Motivation
+[motivation]: #motivation
+
+Indy Plenum can be extended through the use of [ledger plugins](https://github.com/hyperledger/indy-plenum/blob/master/docs/source/plugins.md). Ledger plugins can introduce new transaction types that are either stored on existing ledgers, or on new ledgers created by the plugins. The [Sovrin Network](http://sovrin.org) used this capability to develop a [token plugin for Indy](https://github.com/sovrin-foundation/token-plugin) that stored fee and value transfer transactions on a token ledger. After experimenting with the plugin, [they decided to remove it](https://github.com/sovrin-foundation/sovrin-sip/tree/master/text/5005-token-removal/README.md).
```
WadeBarnes, line 20: The definitive decision on whether or not to remove the plugin has not been made.

text/0162-frozen-ledgers/README.md
```diff
@@ -0,0 +1,139 @@
...
+* The audit ledger will no longer be able to obtain the root hash of any custom ledgers from the plugin.
+* Any authorization rules (auth_rules) on the config ledger that define fees as being necessary to authorize writes to the domain ledger can no longer be handled by the plugin.
+
+The first consequence is intended. This HIPE proposes a feature to address the second consequences to ensure that the network remains functional after the plugin is removed. The last consequence is [treated in a separate HIPE](https://github.com/hyperledger/indy-hipe/tree/master/text/0163-default-fee-handler).
+
+
+## Tutorial
+[tutorial]: #tutorial
+
+If a ledger has been created, but will never be used, it can be frozen. The LEDGERS_FREEZE transaction will record the root hashes of one or more ledgers in the state trie, and perpetually use those hashes when computing audit ledger validity. This has two important advantages:
```
WadeBarnes, line 33:
```suggestion
If a ledger has been created, but will no longer be used, it can be frozen. The LEDGERS_FREEZE transaction will record the root hashes of one or more ledgers in the state trie, and perpetually use those hashes when computing audit ledger validity. This has two important advantages:
```

text/0162-frozen-ledgers/README.md
```diff
@@ -0,0 +1,139 @@
...
+2. The ability to catch up nodes through consensus.
+3. The ability to audit the history of the ledger to prove that there was no tampering.
+
+When a ledger is frozen, the root hash of the deprecated ledger is available to be used by the audit ledger when computing consensus. So the first property is preserved.
+
+During catchup, the transaction validation logic is not executed. This is because we are catching up transactions that have the ledger BLS signature, so we know it was already validated and we don't need to do it again. So the second property is preserved.
+
+If a ledger is added (so its root hash is expected by the audit ledger), but it was never used (so the root hash never changed), then LEDGERS_FREEZE will preserve the third property even when the ledger is removed.
+
+But the third property will not be preserved if there is a history of transactions on a ledger and then the ledger is removed. This is because the LEDGERS_FREEZE transaction only stores the most recent root hash, and not the entire history of root hashes, so that history can not be recreated after the plugin ledger is removed. Instead, the frozen ledger should be retained in its read-only state.
```
WadeBarnes, line 117: Using the Sovrin Ledgers as an example, the token ledger has been used on BuilderNet and StagingNet, but not on MainNet.  Based on this statement it would be safe to delete the token ledger data from MainNet, but not on BuilderNet, or StagingNet.  If the ledger were to be deleted from either BuilderNet, or StagingNet the ability to audit the history of the ledger would fail, but how would that present itself, and how do we protect against deleting a ledger that has been used so we can ensure the historical integrity of the ledgers?

WadeBarnes: Added an issue against the `remove_ledger.py` script; https://github.com/hyperledger/indy-node/issues/1659

text/0163-default-fee-handler/README.md
```diff
@@ -0,0 +1,75 @@
...
+## Summary
+[summary]: #summary
+
+Adding a default fee handler to Indy will make it easier to maintain ledgers with payment plugins.
+
+
+## Motivation
+[motivation]: #motivation
+
+Indy Plenum can be extended through the use of [ledger plugins](https://github.com/hyperledger/indy-plenum/blob/master/docs/source/plugins.md). Ledger plugins can introduce new transaction types that are either stored on existing ledgers, or on new ledgers created by the plugins. The [Sovrin Network](http://sovrin.org) used this capability to develop a [token plugin for Indy](https://github.com/sovrin-foundation/token-plugin) that stored fee and value transfer transactions on a token ledger. After experimenting with the plugin, [they decided to remove it](https://github.com/sovrin-foundation/sovrin-sip/tree/master/text/5005-token-removal/README.md).
```
WadeBarnes, line 21: The definitive decision on whether or not to remove the plugin has not been made.

text/0163-default-fee-handler/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+Indy will also look to the default fee handler to determine the execution logic for fee related transactions, but the default fee handler will not allow any new fee transactions and so doesn't need any specific execution logic. Any future implementation of a payment plugin will need to override the default fee handler with a functional implementation of validation and execution logic.
+
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
+* Historically there were concerns about having payment functionality included in Hyperledger projects, but as distributed payments have become more mainstream, it now seems acceptable to include into Indy generic primitives related to payments.
+* The default fee handler replays transactions without performing validation. This is fine for catching up validator nodes, but an audit of the ledger history would need to take into account the historical validation expected by any removed plugins.
```
WadeBarnes, line 53: How is this accomplished once the plugin is removed?

text/0162-frozen-ledgers/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Motivation
+[motivation]: #motivation
+
+Indy Plenum can be extended through the use of [ledger plugins](https://github.com/hyperledger/indy-plenum/blob/master/docs/source/plugins.md). Ledger plugins can introduce new transaction types that are either stored on existing ledgers, or on new ledgers created by the plugins. The [Sovrin Network](http://sovrin.org) used this capability to develop a [token plugin for Indy](https://github.com/sovrin-foundation/token-plugin) that stored fee and value transfer transactions on a token ledger. After experimenting with the plugin, [they decided to remove it](https://github.com/sovrin-foundation/sovrin-sip/tree/master/text/5005-token-removal/README.md).
+
+Removing the plugin has three important consequences:
+* The data stored on the custom ledger created by the plugin will be deleted.
+* The audit ledger will no longer be able to obtain the root hash of any custom ledgers from the plugin.
+* Any authorization rules (auth_rules) on the config ledger that define fees as being necessary to authorize writes to the domain ledger can no longer be handled by the plugin.
```
swcurran, line 25: I assume this is specific to the token plugin vs. some other arbitrary plugin?

WadeBarnes:
We'll continue the conversation from here (https://github.com/hyperledger/indy-node/issues/1659#issuecomment-785081769) on this PR.

text/0162-frozen-ledgers/README.md
esplinr: Agreed. It was intended as context. I made it more generic:

> After experimenting with the plugin, [a plan was made to remove it].

text/0163-default-fee-handler/README.md
esplinr: Agreed. It was intended as context. I made it more generic:

> After experimenting with the plugin, [a plan was made to remove it].

text/0162-frozen-ledgers/README.md
esplinr: Great suggestion.

text/0162-frozen-ledgers/README.md
esplinr: I added a note to clarify that these are three consequences from removing the Sovrin token plugin.

text/0162-frozen-ledgers/README.md
esplinr: HIPE 0162 addresses this in the Drawbacks section.

> 3. The ability to audit the history of the ledger to prove that there was no tampering.
> … the third property will not be preserved if there is a history of transactions on a ledger and then the  ledger is removed. This is because the LEDGERS_FREEZE transaction only stores the most recent root hash, and not the entire history of root hashes, so that history can not be recreated after the plugin ledger is removed. Instead, the frozen ledger should be retained in its read-only state.

As the HIPE explains at the beginning,

>  The data stored on the custom ledger created by the plugin will be deleted.
> [This] consequence is intended.

The associated SIP makes clear that this is acceptable for the Sovrin test networks because they are not intended to have a permanent history.

text/0163-default-fee-handler/README.md
esplinr: This is discussed in the #drawbacks section of HIPE 0162-frozen-ledgers.

This HIPE doesn't discuss removing ledgers, so I think it is sufficient to state that the historical validation will need to be taken into account.

esplinr:
I believe I have incorporated all the feedback and answered all of the questions.

text/0162-frozen-ledgers/README.md
WadeBarnes: The answers focus more heavily on the current intended use case.  I'm thinking more about the impact when something like this is used in general and it's impact to historical integrity.

To summarize my understanding so far:

Once the the ledger has been frozen and deleted, the ability to audit the history is irrevocably lost.
- How does this end up presenting itself from an operational and network administrative standpoint?
- What are the implications to the network and other ledgers?
- What is the impact to the network?

If the ledger is frozen but the data is retained, can the history of the ledger be audited?
- Can this be accomplished if the plugin that created/managed the ledger is removed?

text/0163-default-fee-handler/README.md
WadeBarnes: Audit of ledger history is discussed, but I'm missing the definitive answer as to whether or not the ledger history can be successfully audited in the case the ledger is frozen, the data is retained, and the plug-in is removed.  If this is possible, how would it be accomplished in practice?

text/0162-frozen-ledgers/README.md
WadeBarnes: Would this then be a fair summary?

Once a ledger is frozen and the associated plug-in is removed, regardless of whether or not you keep the data, the history of the associated ledger(s) cannot be audited.  In order to audit their history you would have to retain the ledger data (not delete the ledger after freezing it) and restore the plugin (or computational equivalent) first.

WadeBarnes: How does a frozen ledger affect new nodes added to the network.  I assume, whether or not the frozen ledgers data was retained, only the hash at the point in time the ledger was frozen will be replicated to any new nodes.  Is this correct?

esplinr: > Would this then be a fair summary?
> Once a ledger is frozen and the associated plug-in is removed, regardless of whether or not you keep the data, the history of the associated ledger(s) cannot be audited. In order to audit their history you would have to retain the ledger data (not delete the ledger after freezing it) and restore the plugin (or computational equivalent) first.

I disagree with the sentiment that the history cannot be audited even if you keep the data. The audit process is the same as before, because audits are done outside the ledger software and independent of what plugins are installed on the ledger. To perform an accurate audit you will need to track the history of the software that was installed so that you can confirm the validation rules that were used for the original ledger writes.

esplinr: > How does a frozen ledger affect new nodes added to the network. I assume, whether or not the frozen ledgers data was retained, only the hash at the point in time the ledger was frozen will be replicated to any new nodes. Is this correct? I think this may already be implied by the fact the frozen ledger will not be included in catch up operations.

I believe that the ledger data is propagated during catch up to the rest of the pool, even for frozen ledgers. This is because validation is not done for transactions that already have a BLS signature. Is that correct @Toktar ?

text/0163-default-fee-handler/README.md
esplinr: I added a paragraph to make this more explicit.

text/0162-frozen-ledgers/README.md
esplinr: I see this note from @anton.denishchenko: "Frozen ledgers are not catched up (this is possible because the LEDGERS_FREEZE transaction is part of the  config ledger, which is applied before additional ledgers)."
Thank you @WadeBarnes for highlighting this misunderstanding. I'll add a note about this topic to the HIPE.

text/0162-frozen-ledgers/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+When a ledger is frozen, the root hash of the deprecated ledger is available to be used by the audit ledger when computing consensus. So the first property is preserved.
+
+During catchup, the transaction validation logic is not executed. This is because we are catching up transactions that have the ledger BLS signature, so we know it was already validated and we don't need to do it again. So the second property is preserved.
+
+If a ledger is added (so its root hash is expected by the audit ledger), but it was never used (so the root hash never changed), then LEDGERS_FREEZE will preserve the third property even when the ledger is removed.
+
+But the third property will not be preserved if there is a history of transactions on a ledger and then the ledger is removed. This is because the LEDGERS_FREEZE transaction only stores the most recent root hash, and not the entire history of root hashes, so that history can not be recreated after the plugin ledger is removed. Instead, the frozen ledger should be retained in its read-only state and new nodes in the pool should manually backup the data from frozen ledgers since they are not included in catchup.
```
WadeBarnes, line 118: `and new nodes in the pool should manually backup the data from frozen ledgers since they are not included in catchup.`

This would be additional administrative overhead and could easily be overlooked or done incorrectly.  It's definitely something that needs to be addressed at some point before too long.  I would think It would be best addressed by the existing catch-up process.

How much effort would be required to support catch-up of frozen ledgers?  The need for this should at least be captured in a ticket so it's not forgotten.

esplinr: It's too much work for that feature without someone planning to use it, but here is the issue:
https://github.com/hyperledger/indy-plenum/issues/1524

PR #166, opened 2024-05-10
chore: updating dependabot file to support Python packages

rajpalc7:


WadeBarnes (approved):
Looks good to me, but I'd like feedback from other maintainers.

